PR #5, opened 2024-06-06
M kovalsky/functionparametersdescriptions

m-kovalsky:


sempy_labs/Lakehouse.py
```diff
@@ -4,14 +4,21 @@
 from pyspark.sql import SparkSession
 from delta import DeltaTable
 from .HelperFunctions import resolve_lakehouse_name
+from typing import List, Optional, Union
 
 def lakehouse_attached():
```
eisber, line 9:
```suggestion
def lakehouse_attached() -> bool:
```

PR #23, opened 2024-07-02
fixed add_time_intel in tom, added is_calculated_table, ran black

m-kovalsky:


src/sempy_labs/tom/_model.py
```diff
@@ -3545,24 +3659,37 @@ def add_time_intelligence(
...
-            if m.Name == measure_name:
-                table_name = m.Parent.Name
-
-        if table_name is None:
-            raise ValueError(f"{icons.red_dot} The '{measure_name}' is not a valid measure in the '{self._dataset}' semantic model within the '{self._workspace}' workspace.")
+        try:
+            table_name = next(
+                m.Parent.Name for m in self.all_measures() if m.Name == measure_name
+            )
+        except:
```
eisber, line 3671: is this exception hit if the measure is not found?

if so, I'd rather have a list created, check if the len is zero and then throw. right now you're masking all other exceptions.

src/sempy_labs/tom/_model.py
```diff
@@ -3545,24 +3659,37 @@ def add_time_intelligence(
...
+            raise ValueError(
+                f"{icons.red_dot} The '{date_table}' table is not a valid date table in the '{self._dataset}' semantic model within the '{self._workspace}' workspace."
+            )
 
         # Extract date key from date table
-        for c in self.all_columns():
-            if c.Parent.Name == date_table and c.IsKey:
-                date_key = c.Name
+        try:
+            date_key = next(
```
eisber, line 3684: same as above, don't use exception to handle not-found.

eisber (changes requested):

eisber (approved):

PR #25, opened 2024-07-03
made changes per flake8, updated tom notebook, fixed functions in tom…

m-kovalsky:
… model


src/sempy_labs/_ai.py
```diff
@@ -367,7 +384,7 @@ def generate_aggs(
                 print(
                     f"{icons.green_dot} '{aggTableName}'[{fromColumn}] -> '{toTable}'[{toColumn}] relationship has been added."
                 )
-            except:
+            except Exception:
```
eisber, line 387: do you want to swallow the exception here? at least I would surface the exception message as part of the print.

src/sempy_labs/migration/_create_pqt_file.py
```diff
@@ -188,16 +194,16 @@ def obj_to_dict(obj):
...
-            types,
-            "{%s}Default" % ns,
-            {"Extension": "json", "ContentType": "application/json"},
-        )
-        default2 = ET.SubElement(
-            types,
-            "{%s}Default" % ns,
-            {"Extension": "pq", "ContentType": "application/x-ms-m"},
-        )
+        # default1 = ET.SubElement(
```
eisber, line 197: do you want to disable this?

src/sempy_labs/migration/_migrate_tables_columns_to_semantic_model.py
```diff
@@ -111,7 +111,7 @@ def migrate_tables_columns_to_semantic_model(
 
                         try:
                             tom.model.Tables[tName]
-                        except:
+                        except Exception:
```
eisber, line 114: is an exception thrown if the table does not exist? can you check if the table exists first, then catching the exception. again this is masking all other errors.

eisber (changes requested):

src/sempy_labs/migration/_create_pqt_file.py
m-kovalsky: it turns out these parameters were not used in the function.

src/sempy_labs/_ai.py
m-kovalsky: I updated the exception message to carry forward.

src/sempy_labs/migration/_migrate_tables_columns_to_semantic_model.py
m-kovalsky: updated this so it doesn't need a try/except clause

eisber (approved):

PR #27, opened 2024-07-04
updated modelbpa to use TOM

m-kovalsky:


src/sempy_labs/tom/_model.py
```diff
@@ -1119,8 +1119,12 @@ def get_annotation_value(self, object, name: str):
         str
             The annotation value.
         """
+        try:
+            value = object.Annotations[name].Value
```
eisber, line 1123: can't you check if name is object.annotations? catch generic exception masks real issues

src/sempy_labs/tom/_model.py
```diff
@@ -1119,8 +1119,12 @@ def get_annotation_value(self, object, name: str):
...
+        except Exception:
+            value = ""
```
eisber, line 1125: why not return None if the value really isn't there. can the value of the annotation by an empty string and/or None?

src/sempy_labs/tom/_model.py
```diff
@@ -2814,6 +2825,9 @@ def records_per_segment(self, object: "TOM.Partition"):
                 object=object, name="Vertipaq_RecordsPerSegment"
             )
 
+        if result == "":
```
eisber, line 2828: I'm wondering if you should follow a similar pattern to dict.get(key, default_value). 
if the annotation or extended property isn't there, return the default value.

eisber (changes requested):

src/sempy_labs/tom/_model.py
m-kovalsky: i made the else clause return None

src/sempy_labs/tom/_model.py
m-kovalsky: i updated the functions to return the following:

```python
return int(result) if result is not None else 0
```

eisber (approved):

PR #30, opened 2024-07-08
made fixes for translations, added support for levels, improved print…

m-kovalsky:
…outs


src/sempy_labs/_translations.py
```diff
@@ -101,6 +101,20 @@ def translate_semantic_model(
...
+
+            for o in tom.all_levels():
+                new_data = {
+                    "Object Type": "Level",
+                    "Name": o.Name,
+                    "TName": o.Name,
+                    "Description": o.Description,
+                    "TDescription": o.Description,
+                    "Display Folder": None,
+                    "TDisplay Folder": None,
```
eisber, line 113: what does T stand for? translation?

src/sempy_labs/_translations.py
```diff
@@ -101,6 +101,20 @@ def translate_semantic_model(
...
+                }
+                dfPrep = pd.concat(
```
eisber, line 115: nit: df_prep

src/sempy_labs/_translations.py
```diff
@@ -330,6 +355,33 @@ def translate_semantic_model(
...
                                         )
+                                for lev in h.Levels:
+                                    if clm == "Name":
+                                        df_filt = df_panda[
+                                            (df_panda["Object Type"] == "Level")
+                                            & (df_panda["Name"] == lev.Name)
+                                        ]
+                                        if len(df_filt) == 1:
+                                            tr = df_filt["translation"].str[i].iloc[0]
+                                            tom.set_translation(
```
eisber, line 366: can you refactor into functions/methods? the code looks quite similar. also the long function makes it hard to follow

eisber (changes requested):

src/sempy_labs/_translations.py
m-kovalsky: @eisber there are 2 columns for each (TName and Name) because of the exclude_characters functionality. Since it may change the object name there are 2 versions of each (one to maintain the original name and one to have the updated name (i.e 'Internet_Sales' and 'Internet Sales'). I can probably make this cleaner.

src/sempy_labs/_translations.py
m-kovalsky: updated

src/sempy_labs/_translations.py
m-kovalsky: made 2 subfunctions within this function to greatly simplify the code as recommended.

src/sempy_labs/tom/_model.py
```diff
@@ -1848,7 +1895,7 @@ def hybrid_tables(self):
                 if any(p.Mode == TOM.ModeType.DirectQuery for p in t.Partitions):
                     yield t
 
-    def date_tables(self):
+    def all_date_tables(self):
```
eisber, line 1898: if you rename methods (which breaks the API) you need to increase the major version

src/sempy_labs/migration/_refresh_calc_tables.py
```diff
@@ -121,9 +122,11 @@ def refresh_calc_tables(dataset: str, workspace: Optional[str] = None):
                                 f"{icons.green_dot} Calculated table '{tName}' has been refreshed as the '{delta_table_name.lower()}' table in the lakehouse."
                             )
                         except Exception as e:
```
eisber, line 124: for a separate PR, we should do a round of exception reviewing. I think there's still a bunch of cases, where you can check on a condition before hand and avoid just catching the exception

eisber (approved):

PR #41, opened 2024-07-24
cleaned up f strings

m-kovalsky:


src/sempy_labs/_generate_semantic_model.py
```diff
@@ -126,7 +126,7 @@ def create_semantic_model_from_bim(
         lro_wait=True,
     )
 
-    if response.status_code != 200:
+    if response.status_code not in [200, 201]:
```
eisber, line 129: I'm surprised that you get both? don't have to change. just clarification.

src/sempy_labs/_helper_functions.py
```diff
@@ -286,26 +286,34 @@ def get_direct_lake_sql_endpoint(dataset: str, workspace: Optional[str] = None)
...
-    dfE_filt = dfE[dfE["Name"] == "DatabaseQuery"]
-    expr = dfE_filt["Expression"].iloc[0]
+    with connect_semantic_model(
+        dataset=dataset, readonly=True, workspace=workspace
+    ) as tom:
+        sqlEndpointId = None
+        for e in tom.model.Expressions:
+            if e.Name == "DatabaseQuery":
+                expr = e.Expression
+                matches = re.findall(r'"([^"]*)"', expr)
```
eisber, line 310: you also want to find "" ? (so no characters in between quotes?

eisber (approved):

src/sempy_labs/_generate_semantic_model.py
m-kovalsky: i think it's 201 for this one but generally 201 or 200 are successes.

src/sempy_labs/_helper_functions.py
m-kovalsky: it will just be a guid between the quotes

PR #43, opened 2024-07-25
added language translation capability for run_model_bpa

m-kovalsky:


src/sempy_labs/_model_bpa.py
```diff
@@ -89,6 +99,69 @@ def run_model_bpa(
         rules["Severity"].replace("Error", "\u274C", inplace=True)
         rules["Severity"].replace("Info", "ℹ️", inplace=True)
 
+        # Translate
+        if language is not None:
+            rules_temp = rules.copy()
```
eisber, line 104: The rule names are not going to change per customer, right? shouldn't we just run the translation once and commit the files?

m-kovalsky: done - used .po file for 20 language support.

src/sempy_labs/_model_bpa.py
```diff
@@ -65,9 +74,44 @@ def run_model_bpa(
...
+        "ta-IN",
+        "th-TH",
+        "zu-ZA",
+        "am-ET",
+        "ar-AE",
+    ]
 
     workspace = fabric.resolve_workspace_name(workspace)
 
+    if language is not None and language not in language_list:
```
eisber, line 112: I think gettext supports fallback if the country spec doesn't match (e.g. de-AT -> de-DE)?

src/sempy_labs/_model_bpa.py
```diff
@@ -80,7 +124,84 @@ def run_model_bpa(
 
         dep = get_model_calc_dependencies(dataset=dataset, workspace=workspace)
 
-        if rules is None:
+        # Translations
+        if language is not None and rules is None and language in language_list:
+            # Use the polib to get the translations without using spark
+            rules = model_bpa_rules(
```
eisber, line 130: move the blocks in if ... elseif ... else ... into separate private functions to keep the top-level readable.

src/sempy_labs/_bpa_translation/_translations_category.po
```diff
@@ -0,0 +1,6403 @@
+#
+msgid ""
+msgstr ""
+"Language: en\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+
+# it-IT
+msgid "Do not use floating point data types"
```
eisber, line 10: something seems off here? message is way to short and the reverse translation doesn't seem to make sense?

eisber (changes requested):

src/sempy_labs/_bpa_translation/_translations_category.po
m-kovalsky: there are 3 .po files, one for rule name, category, description. They all map via rule name so the msgid is the Rule Name and the msgstr is the translation (either rule name, category or description).

src/sempy_labs/_model_bpa.py
m-kovalsky: done

src/sempy_labs/_model_bpa.py
m-kovalsky: I added a function (map_language) which does this.

src/sempy_labs/_model_bpa.py
```diff
@@ -107,10 +107,32 @@ def run_model_bpa(
         "ar-AE",
     ]
 
+    # Map languages to the closest language (first 2 letters matching)
+    def map_language(language, language_list):
+
+        mapped = False
+
+        if language in language_list:
+            mapped is True
```
eisber, line 116: what's the purpose of this statement?

src/sempy_labs/_model_bpa.py
```diff
@@ -186,25 +205,41 @@ def run_model_bpa(
...
             for clm in columns:
-                rules = rules.drop([clm], axis=1)
-                rules = rules.rename(columns={f"{clm}Translated": clm})
-        elif rules is None:
+                rule_file = rule_file.drop([clm], axis=1)
+                rule_file = rule_file.rename(columns={f"{clm}Translated": clm})
+
+            return rule_file
+
+        translated = False
```
eisber, line 228: same here.

eisber (approved):

src/sempy_labs/_model_bpa.py
m-kovalsky: i only want it to do the mapping if the language doesn't exist in the language list. otherwise it could map it to None.

src/sempy_labs/_model_bpa.py
m-kovalsky: this identifies if the rules have been translated. it makes the logic simpler in the lines below this.

PR #63, opened 2024-08-06
M kovalsky/generatedlsemanticmodel

m-kovalsky:


src/sempy_labs/directlake/_dl_helper.py
```diff
@@ -0,0 +1,175 @@
...
+        )
+
+    create_blank_semantic_model(dataset=dataset, workspace=workspace)
+
+    start_time = datetime.datetime.now()
+    timeout = datetime.timedelta(minutes=1)
+    success = False
+    expression_name = "DatabaseQuery"
+
+    while not success:
```
eisber, line 145: can you move the timeout into the while condition? makes it easier to read. also why use success = True if you break later? why not always break?

eisber (changes requested):

src/sempy_labs/directlake/_dl_helper.py
m-kovalsky: success=True stops the while loop. if i don't put it there it will spin for a minute and keep adding and readding the same table repeatedly.

m-kovalsky: added decorator function

src/sempy_labs/_helper_functions.py
```diff
@@ -670,3 +673,21 @@ def resolve_workspace_capacity(workspace: Optional[str] = None) -> Tuple[UUID, s
...
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            start_time = datetime.datetime.now()
+            timeout = datetime.timedelta(minutes=1)
+            while datetime.datetime.now() - start_time <= timeout:
+                try:
+                    return func(*args, **kwargs)
+                except Exception:
+                    time.sleep(sleep_time)
```
eisber, line 688: can be for the next pr: remember the last exception and pass along w/ timeouterror. e.g. what happens if this fails due to some other reason

src/sempy_labs/directlake/_dl_helper.py
```diff
@@ -137,39 +136,32 @@ def generate_direct_lake_semantic_model(
...
-                            table_name=t,
-                            column_name=lakeCName,
-                            source_column=lakeCName,
-                            data_type=dt,
-                        )
-        except Exception:
-            if datetime.datetime.now() - start_time > timeout:
-                break
-            time.sleep(1)
+    @retry(sleep_time=1, timeout_error_message="Function timed out after 1 minute")
```
eisber, line 141: checking every second seems a bit aggressive. maybe do 5?
also the error message is probably better to state: dyn_create_model function failed after timeout. you can generate the message by doing func.__name__

eisber (approved):

src/sempy_labs/directlake/_dl_helper.py
m-kovalsky: i think 5 is too long. i tried that before and it makes it take longer on average.

PR #65, opened 2024-08-07
Update _report_functions.py to work for paginated reports with no vis…

jugi92:
…uals or pages given.

if page and visual are empty the function would fail because it would still run `list_report_pages` or `list_report_visuals`. Moved that into the if clause to check it properly.


m-kovalsky:
Thanks. Updated in PR #190.

m-kovalsky:
Added as part of [0.8.3](https://github.com/microsoft/semantic-link-labs/releases/tag/0.8.3). Thanks for your feedback!

PR #84, opened 2024-08-16
fixed to support pagination

m-kovalsky:


src/sempy_labs/lakehouse/_get_lakehouse_tables.py
```diff
@@ -92,99 +94,108 @@ def get_lakehouse_tables(
...
+            if not extended:
+                new_data = {
+                    "Workspace Name": workspace,
+                    "Lakehouse Name": lakehouse,
+                    "Table Name": tName,
+                    "Format": tFormat,
+                    "Type": tType,
+                    "Location": tLocation,
+                }
                 df = pd.concat(
```
bo-zhou-ms, line 114: Consider collecting all dataframes into a list first and then concat the whole list for one time? This should be faster than do it one by one. E.g. 
```python
dfs = []
for r in responses:
        for i in r.get("data", []):
            ...
            dfs.append(pd.DataFrame(new_data, index=[0]))
df = pd.concat(dfs, ignore_index=True)
```

Could also update occurrences of concating in other functions.

src/sempy_labs/_list_functions.py
```diff
@@ -1316,21 +1345,18 @@ def list_workspace_role_assignments(workspace: Optional[str] = None) -> pd.DataF
...
-            "Role Name": role_name,
-            "Type": user_type,
-            "User Email": user_email,
-        }
-        df = pd.concat([df, pd.DataFrame(new_data, index=[0])], ignore_index=True)
+    for r in responses:
+        for i in r.get("value", []):
+            principal = i.get("principal", {})
+            new_data = {
+                "User Name": principal.get("displayName"),
```
bo-zhou-ms, line 1354: Not much for this PR but maybe we could consider adding a function to parse data from response (or responses together with your `pagination`) to DataFrame since this piece of code is reused in multiple places.

src/sempy_labs/lakehouse/_get_lakehouse_tables.py
m-kovalsky: updated

PR #106, opened 2024-08-30
added list_dataflows

m-kovalsky:


src/sempy_labs/_dataflows.py
```diff
@@ -0,0 +1,133 @@
...
+    print(
+        f"{icons.green_dot} The '{dataflow_storage_account}' dataflow storage account has been assigned to the '{workspace}' workspacce."
+    )
+
+
+def list_dataflow_storage_accounts() -> pd.DataFrame:
+    """
+    Shows the accessible dataflow storage accounts.
+
+    Parameters
```
eisber, line 102: remove section if there are no parameters

eisber (changes requested):

PR #112, opened 2024-09-01
only set annotations if the object exists

m-kovalsky:


src/sempy_labs/tom/_model.py
```diff
@@ -2737,57 +2737,69 @@ def set_vertipaq_annotations(self):
...
-                self.set_annotation(object=c, name="Vertipaq_DataSize", value=dataSize)
-                self.set_annotation(
-                    object=c, name="Vertipaq_DictionarySize", value=dictSize
-                )
-                self.set_annotation(
-                    object=c, name="Vertipaq_HierarchySize", value=hierSize
-                )
-                self.set_annotation(object=c, name="Vertipaq_Cardinality", value=card)
+                if len(dfC_filt) > 0:
+                    totalSize = str(dfC_filt["Total Size"].iloc[0])
```
eisber, line 2752: nit: just to reduce the clutter:

row = dfC_filter.iloc[0]

totalSize = str(row["Total Size"])

eisber (approved):

src/sempy_labs/tom/_model.py
m-kovalsky: updated this accordingly

Issue #116, opened 2024-09-03
get_lakehouse_tables() returns error when lakehouse has no tables


joshrodgersKA:
When running get_lakehouse_tables() on a Lakehouse with no tables, the function returns the below error:

`File /nfs4/pyenv-87fb69ad-6cef-4dbb-ad92-d1db14b53e2d/lib/python3.10/site-packages/sempy_labs/lakehouse/_get_lakehouse_tables.py:111, in get_lakehouse_tables(lakehouse, workspace, extended, count_rows, export)
    102         new_data = {
    103             "Workspace Name": workspace,
    104             "Lakehouse Name": lakehouse,
   (...)
    108             "Location": i.get("location"),
    109         }
    110         dfs.append(pd.DataFrame(new_data, index=[0]))
--> 111 df = pd.concat(dfs, ignore_index=True)
    113 if extended:
    114     sku_value = get_sku_size(workspace)

File ~/cluster-env/trident_env/lib/python3.10/site-packages/pandas/core/reshape/concat.py:372, in concat(objs, axis, join, ignore_index, keys, levels, names, verify_integrity, sort, copy)
    369 elif copy and using_copy_on_write():
    370     copy = False
--> 372 op = _Concatenator(
    373     objs,
    374     axis=axis,
    375     ignore_index=ignore_index,
    376     join=join,
    377     keys=keys,
    378     levels=levels,
    379     names=names,
    380     verify_integrity=verify_integrity,
    381     copy=copy,
    382     sort=sort,
    383 )
    385 return op.get_result()

File ~/cluster-env/trident_env/lib/python3.10/site-packages/pandas/core/reshape/concat.py:429, in _Concatenator.__init__(self, objs, axis, join, keys, levels, names, ignore_index, verify_integrity, copy, sort)
    426     objs = list(objs)
    428 if len(objs) == 0:
--> 429     raise ValueError("No objects to concatenate")
    431 if keys is None:
    432     objs = list(com.not_none(*objs))

ValueError: No objects to concatenate`

m-kovalsky:
Thanks for reporting.

m-kovalsky:
This is fixed in [0.7.3](https://github.com/microsoft/semantic-link-labs/releases/tag/0.7.3).

PR #117, opened 2024-09-03
fixed issue 116 for pagination

m-kovalsky:


m-kovalsky:
#116

src/sempy_labs/lakehouse/_get_lakehouse_tables.py
```diff
@@ -108,7 +108,9 @@ def get_lakehouse_tables(
                 "Location": i.get("location"),
             }
             dfs.append(pd.DataFrame(new_data, index=[0]))
-    df = pd.concat(dfs, ignore_index=True)
+
+    if dfs:
```
eisber, line 112: should you maybe return straight from here if you don't have rows?
are the extended/count_rows/... paths going to do what you want?

eisber (changes requested):

src/sempy_labs/lakehouse/_get_lakehouse_tables.py
m-kovalsky: updated accordingly

src/sempy_labs/lakehouse/_get_lakehouse_tables.py
```diff
@@ -96,6 +95,9 @@ def get_lakehouse_tables(
 
     responses = pagination(client, response)
 
+    if not responses[0].get("data"):
```
eisber, line 98: possible that response itself has no elements?

eisber (changes requested):

eisber (approved):

PR #121, opened 2024-09-04
M kovalsky/dynamicdocumentation

m-kovalsky:


.github/workflows/_generate_func_examples.py
```diff
@@ -0,0 +1,97 @@
...
+                params = [param for param_name, param in sig.parameters.items() if param_name not in ['kwargs', 'self']]
+                param_count = len(params)
+                for param_name, param in sig.parameters.items():
+                    is_optional = False
+                    if param_name not in ['kwargs', 'self']:
+                        param_value = ''
+                        if param.default != inspect.Parameter.empty:
+                            param_value = param.default
+                            is_optional = True
+                        elif param_name == 'dataset':
```
eisber, line 58: can you move to a dictionary to make this less clutter?

.github/workflows/_generate_func_examples.py
```diff
@@ -0,0 +1,97 @@
...
+    TOMWrapper: 'tom',
+}
+
+link_prefix = "https://semantic-link-labs.readthedocs.io/en/stable/"
+tab = '    '
+skip_functions = ['connect_semantic_model', '__init__', 'close']
+
+markdown_example = '## Function Examples\n'
+# Function Examples
+for d, d_alias in dirs.items():
```
eisber, line 26: can you reduce the nesting a bit by maybe building up a list first?

from what I can see you should be able to a do single list comprehension with loops + ifs?

.github/workflows/_generate_readme.py
```diff
@@ -0,0 +1,258 @@
...
+functions = {}
+for d, d_alias in dirs.items():
+    d_name = d.__name__   
+    for attr_name in dir(d):
+        attr = getattr(d, attr_name)
+        if inspect.isfunction(attr):
+            if attr_name not in skip_functions:
+                sig = inspect.signature(attr)
+                category = 'General'
+                if d_alias == 'migration':
```
eisber, line 90: move to dictionary or into the signature using decorators?

PR #146, opened 2024-09-15
added capacity functions

m-kovalsky:


src/sempy_labs/_capacities.py
```diff
@@ -0,0 +1,541 @@
...
+        The name of the region in which the capacity will be created.
+    sku : str
+        The `sku size <https://azure.microsoft.com/pricing/details/microsoft-fabric/>`_ of the Fabric capacity.
+    admin_members : str | List[str]
+        The email address(es) of the admin(s) of the Fabric capacity.
+    tags: dict, default=None
+        Tag(s) to add to the capacity. Example: {'tagName': 'tagValue'}.
+    """
+
+    from azure.mgmt.resource import ResourceManagementClient
```
eisber, line 65: I'm wondering if a deep link/"deploy to azure" experience would make it easier for the customer?
can you provide a link that the customer needs to click on, which opens a new browser window that lets him authenticate?
setting up service principal, keyvault,... seems like a huge pain.

eisber (approved):

Issue #133, opened 2024-09-12
warehouse_connection - DDL Queries?


joshrodgersKA:
I'm trying out the new warehouse_connection function. When running a DDL query, it seems to hang and never complete. The same query when ran in the Warehouse UI works instantly. If this is an issue with the Warehouse rather than the semantic-link package - let me  know and I'll open a ticket with support instead.

Thanks! 

![image](https://github.com/user-attachments/assets/775a4a78-e391-4955-b7ed-ebac7c8d259b)

Here's some sample code that I cannot get to work.

```
warehouse_connection = sempy_labs.ConnectWarehouse("Oracle_Silver")
sql = """
CREATE TABLE test_table (
    id INT ,
    name VARCHAR(255)
)
"""
warehouse_connection.query(sql)
```

m-kovalsky:
Would you try using the syntax shown in this notebook:

https://github.com/microsoft/semantic-link-labs/blob/main/notebooks/Warehouse.ipynb
________________________________
From: Josh-Rodgers ***@***.***>
Sent: Thursday, September 12, 2024 5:23:13 AM
To: microsoft/semantic-link-labs ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [microsoft/semantic-link-labs] warehouse_connection - DDL Queries? (Issue #133)


I'm trying out the new warehouse_connection function. When running a DDL query, it seems to hang and never complete. The same query when ran in the Warehouse UI works instantly. If this is an issue with the Warehouse rather than the semantic-link package - let me know and I'll open a ticket with support instead.

Thanks!

image.png (view on web)<https://github.com/user-attachments/assets/775a4a78-e391-4955-b7ed-ebac7c8d259b>

Here's some sample code that I cannot get to work.

warehouse_connection = sempy_labs.ConnectWarehouse("Oracle_Silver")
sql = """
CREATE TABLE test_table (
    id INT ,
    name VARCHAR(255)
)
"""
warehouse_connection.query(sql)

joshrodgersKA:
Tried this:

```
warehouse_name = "Oracle_Silver"
workspace_name = "KDW-Dev"
with sempy_labs.ConnectWarehouse(warehouse=warehouse_name, workspace=workspace_name) as sql:
    df = sql.query("CREATE TABLE dbo.test_table1 (id INT, name VARCHAR(255))")
```

Got the below error: 

---------------------------------------------------------------------------
ProgrammingError                          Traceback (most recent call last)
[...]

PR #148, opened 2024-09-15
M kovalsky/warehousetsql

m-kovalsky:


m-kovalsky:
#133

src/sempy_labs/_sql.py
```diff
@@ -59,29 +59,51 @@ def __init__(
...
-            )
+            for sql_query in sql:
+                cursor.execute(sql_query)
+
+                # Commit for non-select queries (like CREATE, INSERT, etc.)
+                if not cursor.description:
+                    self.connection.commit()
+                else:
+                    # Fetch and append results for queries that return a result set
+                    result = pd.DataFrame.from_records(
```
eisber, line 94: that behavior seems a bit strange. why would I run multiple SQL queries that have the same schema?

if there are multiple sql queries, I'd suggest to return a list of data frames.

eisber (changes requested):

eisber (approved):

PR #155, opened 2024-09-17
added functionality for connectlakehouse

m-kovalsky:


src/sempy_labs/_sql.py
```diff
@@ -89,22 +98,16 @@ def query(
             for sql_query in sql:
                 cursor.execute(sql_query)
 
-                # Commit for non-select queries (like CREATE, INSERT, etc.)
```
eisber, line 92: why remove the comments?

m-kovalsky: added them back in

eisber (approved):

Issue #165, opened 2024-09-23
Automatic conversion of field names to 'business friendly' field names


aventius-software:
**Is your feature request related to a problem? Please describe.**
I've got a semantic model with loads of fields, all named as they are in the source tables, so maybe Pascal case or snake case, I currently have to rename them all manually to more business friendly names in the semantic model - which is a real pain.

**Describe the solution you'd like**
In the old days of SSAS multidimensional cube projects in Visual Studio years ago it would do this for you... When creating a model from a data source it recognised word separation in the original data source field name (i.e. snake case or Pascal case named columns). It then auto converted them to nice 'business friendly' for you in your cube project without you needing to do anything! So for example, original column names like 'ThisIsAField' or 'this_is_a_field' would be translated into 'This Is A Field'. Would be great if there was something that would do this in Fabric.

**Describe alternatives you've considered**
Currently just manually doing this, not aware of alternatives, unless there's some way of scripting the change...

**Additional context**
N/A

m-kovalsky:
You can use the TOMWrapper to loop through your model objects (tables, columns, measures etc) and use RegEx to convert and rename snake/pascal case to friendly case.
________________________________
From: Rocket Porg ***@***.***>
Sent: Monday, September 23, 2024 11:01:39 AM
To: microsoft/semantic-link-labs ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [microsoft/semantic-link-labs] Automatic conversion of field names to 'business friendly' field names (Issue #165)


Is your feature request related to a problem? Please describe.
I've got a semantic model with loads of fields, all named as they are in the source tables, so maybe Pascal case or snake case, I currently have to rename them all manually to more business friendly names in the semantic model - which is a real pain.

Describe the solution you'd like
In the old days of SSAS multidimensional cube projects in Visual Studio years ago it would do this for you... When creating a model from a data source it recognised word separation in the original data source field name (i.e. snake case or Pascal case named columns). It then auto converted them to nice 'business friendly' for you in your cube project without you needing to do anything! So for example, original column names like 'ThisIsAField' or 'this_is_a_field' would be translated into 'This Is A Field'. Would be great if there was something that would do this in Fabric.

Describe alternatives you've considered
Currently just manually doing this, not aware of alternatives, unless there's some way of scripting the change...

Additional context
N/A

m-kovalsky:
[...]

PR #168, opened 2024-09-23
added convert_to_friendly_case function

m-kovalsky:
#165


src/sempy_labs/_helper_functions.py
```diff
@@ -1048,3 +1048,25 @@ def get_azure_token_credentials(
     }
 
     return token, credential, headers
+
+
+def convert_to_friendly_case(text: str) -> str:
```
eisber, line 1053: can you add a unit test for that one? regex is always hard to read.

eisber (changes requested):

eisber (approved):

src/sempy_labs/_helper_functions.py
m-kovalsky: done

PR #209, opened 2024-10-15
removed region from trial capacity migration

m-kovalsky:


src/sempy_labs/_capacities.py
```diff
@@ -521,20 +521,41 @@ def update_fabric_capacity(
...
+    payload["properties"] = get_json["properties"]
+
     if sku is not None:
-        payload["sku"] = {"name": sku, "tier": "Fabric"}
+        payload["sku"]["name"] = sku
     if admin_members is not None:
-        payload["properties"] = {"administration": {"members": [admin_members]}}
-
-    payload = _add_sll_tag(payload, tags)
+        payload["properties"]["administration"]["members"] = admin_members
```
eisber, line 546: can it be that administration and members don't exist?

m-kovalsky: Updated. However, if admin members aren't specified, the payload 'properties' will be taken from the existing value (see line 541)

eisber (approved):

PR #218, opened 2024-10-16
altText support for field values and measures in addition to literals

stephbruno:
Added 2 helper functions to get expressions as well as literals. Changed the AltText column in the list_visulas function to use the new functions.


stephbruno:

m-kovalsky:
This is going into 0.8.7 with PR #298

Issue #233, opened 2024-10-28
get_direct_lake_guardrails() returns Fabric SKU not Fabric SKUs as a column name.


MitchSS:
get_direct_lake_guardrails() returns Fabric SKU not Fabric SKUs as a column name. Therefore, get_lakehouse_tables() fails when using extended=True.

get_directlake_guardrails_for_sku is also affected.

m-kovalsky:
Fixed in [0.8.4](https://github.com/microsoft/semantic-link-labs/releases/tag/0.8.4).

PR #234, opened 2024-10-28
PR - Updated Column name from Fabric SKUs to Fabric SKUs

MitchSS:
Resolves issue #233


m-kovalsky:
Thanks - I already fixed this in another PR to be checked in.

PR #236, opened 2024-10-29
Update _helper_functions.py

itsnotaboutthecell:
Updated resolve_workspace_name_and_id to include resolve_workspace_name for a UUID submission.


m-kovalsky:
adding the '_is_valid_uuid' helper function which is more robust than this code.

PR #237, opened 2024-10-29
Update _helper_functions.py

itsnotaboutthecell:
Added resolve_dataflow_name_and_id to return dataflow name and ID information.


m-kovalsky:
#264 will take care of this

Issue #246, opened 2024-11-01
Vertipaq Analyzer export to zip


danmeissner:
Exporting to zip when running the .vertipaq_analyzer(dataset, workspace, export) method is throwing the following error.  Exporting to table works fine.  This is repeatable across tenants, workspaces.

```

--------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[22], line 1
----> 1 labs.vertipaq_analyzer(dataset = dataset_name, workspace = workspace_name, export = 'zip')

File /nfs4/pyenv-d5950a54-a991-4136-b5b5-beb174d95096/lib/python3.11/site-packages/sempy/_utils/_log.py:310, in mds_log.<locals>.get_wrapper.<locals>.log_decorator_wrapper(*args, **kwargs)
    307     raise
    309 try:
--> 310     result = func(*args, **kwargs)
    312     # The invocation for get_message_dict moves after the function
    313     # so it can access the state after the method call
    314     message.update(extractor.get_completion_message_dict(result, arg_dict))

File /nfs4/pyenv-d5950a54-a991-4136-b5b5-beb174d95096/lib/python3.11/site-packages/sempy_labs/_vertipaq.py:610, in vertipaq_analyzer(dataset, workspace, export, read_stats_from_data, **kwargs)
    608 for fileName, df in dataFrames.items():
    609     filePath = os.path.join(subFolderPath, f"{fileName}{ext}")
--> 610     df.to_csv(filePath, index=False)
    612 # Create a zip file and add CSV files to it
    613 with zipfile.ZipFile(zipFilePath, "w") as zipf:

AttributeError: 'Styler' object has no attribute 'to_csv'

```
![Image](https://github.com/user-attachments/assets/99f3f750-773d-4838-b79c-0907294e2766)

danmeissner:
Is there a library that I am not installing or missing?

m-kovalsky:
Fixed in [0.8.5](https://github.com/microsoft/semantic-link-labs/releases/tag/0.8.5).

PR #248, opened 2024-11-03
fixed issue 246 (vertipaq)

m-kovalsky:
#246

### DIFF
--- a/src/sempy_labs/_vertipaq.py
+++ b/src/sempy_labs/_vertipaq.py
@@ -412,18 +412,19 @@ def vertipaq_analyzer(
     export_Model = dfModel.copy()
 
     def _style_columns_based_on_types(dataframe: pd.DataFrame, column_type_mapping):
-
-        format_mapping = {
-            "int": "{:,}",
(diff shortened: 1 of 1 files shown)


src/sempy_labs/_vertipaq.py
```diff
@@ -979,7 +980,8 @@ def import_vertipaq_analyzer(folder_path: str, file_name: str):
     dfs = {}
     for file_name in zip_ref.namelist():
         df = pd.read_csv(extracted_dir + "/" + file_name)
-        dfs[file_name] = df
+        df_name = file_name[:-4]
```
eisber, line 983: I assume you're trimming of the file extension? better to use [pathlib.Path.stem](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem)

eisber (changes requested):

src/sempy_labs/_vertipaq.py
m-kovalsky: fixed

eisber (approved):

PR #257, opened 2024-11-05
added mirroring functions

m-kovalsky:


src/sempy_labs/_mirrored_databases.py
```diff
@@ -0,0 +1,436 @@
...
+    for r in responses:
+        for v in r.get("value", []):
+            prop = v.get("properties", {})
+            sql = prop.get("sqlEndpointProperties", {})
+            new_data = {
+                "Mirrored Database Name": v.get("displayName"),
+                "Mirrored Database Id": v.get("id"),
+                "Description": v.get("description"),
+                "OneLake Tables Path": (
+                    prop.get("oneLakeTablesPath") if prop is not None else None
```
eisber, line 64: I don't think prop can be None

src/sempy_labs/_mirrored_databases.py
```diff
@@ -0,0 +1,436 @@
...
+    )
+
+    for r in responses:
+        for v in r.get("data", []):
+            m = v.get("metrics", {})
+            new_data = {
+                "Source Schema Name": v.get("sourceSchemaName"),
+                "Source Table Name": v.get("sourceTableName"),
+                "Status": v.get("status"),
+                "Processed Bytes": m.get("processedBytes") if m is not None else None,
```
eisber, line 254: m can't be None?

eisber (changes requested):

src/sempy_labs/_mirrored_databases.py
m-kovalsky: fixed

src/sempy_labs/_mirrored_databases.py
m-kovalsky: fixed

eisber (approved):

PR #264, opened 2024-11-10
added list_upstream_dataflows

m-kovalsky:


src/sempy_labs/_dataflows.py
```diff
@@ -132,3 +134,106 @@ def list_dataflow_storage_accounts() -> pd.DataFrame:
...
+                    "Dataflow Id": dataflow_id,
+                    "Workspace Name": workspace_name,
+                    "Workspace Id": workspace_id,
+                    "Upstream Dataflow Name": tgt_dataflow_name,
+                    "Upstream Dataflow Id": tgt_dataflow_id,
+                    "Upstream Workspace Name": tgt_workspace_name,
+                    "Upstream Workspace Id": tgt_workspace_id,
+                }
+
+                nonlocal df
```
eisber, line 208: what's nonlocal here?
can you just pass df and return?

m-kovalsky: updated

eisber (approved):

PR #278, opened 2024-11-14
fixed issue 274, 277

m-kovalsky:


src/sempy_labs/_model_bpa_bulk.py
```diff
@@ -54,6 +55,8 @@ def run_model_bpa_bulk(
         }
     """
 
+    warnings.simplefilter(action='ignore', category=FutureWarning)
```
eisber, line 58: add a comment which future warning you'll filter out

eisber (changes requested):

src/sempy_labs/_model_bpa_bulk.py
```diff
@@ -55,7 +55,8 @@ def run_model_bpa_bulk(
         }
     """
 
-    warnings.simplefilter(action='ignore', category=FutureWarning)
+    # Removing this warning: FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
+    warnings.simplefilter(action="ignore", category=FutureWarning)
```
eisber, line 59: I think you need to fix this. some DF seems all empty and you're concatenating it. can't you simply check for df.empty and not concat?

eisber (changes requested):

eisber (approved):

PR #284, opened 2024-11-17
added connection and gateway apis

m-kovalsky:


src/sempy_labs/_connections.py
```diff
@@ -143,49 +234,93 @@ def list_item_connections(
...
+
+    responses = pagination(client, response)
+
+    records = []
+    for r in responses:
+        for v in r.get("value", []):
+            records.append(
+                {
+                    "Connection Type": v.get("type"),
+                    "Creation Method": v["creationMethods"][0]["name"],
```
eisber, line 270: why not get()?

src/sempy_labs/_connections.py
```diff
@@ -194,157 +329,108 @@ def create_connection_cloud(
...
-    df = pd.concat([df, pd.DataFrame(new_data, index=[0])], ignore_index=True)
-
-    df["Skip Test Connection"] = df["Skip Test Connection"].astype(bool)
 
-    return df
+    print(f"{icons.green_dot} The '{name}' on-prem connection has been created.")
 
 
-def create_connection_vnet(
+def create_vnet_connection(
```
eisber, line 395: are these public functions? if so, docstring?

m-kovalsky: they weren't public as they weren't in the init file

m-kovalsky: done

src/sempy_labs/_connections.py
m-kovalsky: true. making this a private function for now as it's not fully ready.

eisber (approved):

PR #287, opened 2024-11-18
ecotte/authenticationModule

ecotte:
Added the authentication Module for Service Principal API calls.

Objects added:

- ServicePrincipalTokenProviderWithKeyVault: Used to get the Token Provider with Key Vault information.
- ServicePrincipalTokenProvider: Used to get the Token Provider with direct Service Principal information. RECOMMENDED FOR TESTING PUPOSE ONLY.

### DIFF
--- a/src/sempy_labs/_authentication.py
+++ b/src/sempy_labs/_authentication.py
@@ -3,24 +3,21 @@ from sempy.fabric._token_provider import TokenProvider
 import notebookutils
 from azure.identity import ClientSecretCredential
 
-class ServicePrincipalTokenProvider(TokenProvider):   
+
+class ServicePrincipalTokenProvider(TokenProvider):
     """
     Implementation of the sempy.fabric.TokenProvider to be used with Service Principal.
 
     For more information on Service Principal check `Application and service principal objects in Microsoft Entra ID <https://learn.microsoft.com/en-us/entra/identity-platform/app-objects-and-service-principals?tabs=browser#service-principal-object>`_
-
     """
 
-    def __init__(self,credential:ClientSecretCredential):
-
+    def __init__(self, credential: ClientSecretCredential):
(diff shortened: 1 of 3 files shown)


src/sempy_labs/_authentication.py
```diff
@@ -0,0 +1,145 @@
...
+        else:
+            raise NotImplementedError
+
+class ServicePrincipalTokenProvider(TokenProvider):   
+    """
+    Implementation of the sempy.fabric.TokenProvider to be used with Service Principal providing the SP properties.
+
+    For more information on Serice Principal check `Application and service principal objects in Microsoft Entra ID <https://learn.microsoft.com/en-us/entra/identity-platform/app-objects-and-service-principals?tabs=browser#service-principal-object>`_
+
+    ***USE THIS ONE ONLY FOR TEST PURPOSE. FOR PRODUCTION WE RECOMEND THE ServicePrincipalTokenProviderWithKeyVault***
```
eisber, line 88: typo on recommend.
I'd suggest to use warnings lib

src/sempy_labs/_authentication.py
```diff
@@ -0,0 +1,145 @@
...
+            Service Principal App Client ID.
+        client_secret : str
+            Service Principal Secret.
+
+        Returns
+        -------
+        sempy.fabric.TokenProvider
+            Token provider to be used with FabricRestClient or PowerBIRestClient.
+        """
+        self.tenant_id = tenant_id
```
eisber, line 125: can you refactor? the code looks very similar to the other class.

eisber (changes requested):

src/sempy_labs/_authentication.py
ecotte: corrected

PR #293, opened 2024-11-19
ecotte/adminFunctions

ecotte:
Improved the admin functions to support more than 5k workspaces and accept Name or ID in workspaces and capacities.

Functions changed:
- list_capacities: accepts a Capacity ID or Name to filter
- list_workspaces now uses the Fabric API to return any number of workspaces
- list_workspaces now can be filtered with:
  - top: Get only the top N results
  - skip: Skips the top N results
  - capacity: Filters capacity ID or Name
  - workspace: Filters workspace ID or Name
  - workspace_state: If the workspace is Active or Inactive
  - workspace_type: If it´s a workspace or personal workspace
- list_workspaces had the "filter" parameter deprecated
- assign_workspaces_to_capacity: improved to migrate any number of workspaces
- assign_workspaces_to_capacity: improved to accept capacity ID or Name
- assign_workspaces_to_capacity: improved to accept in the workspace list any mix of ID or Names
- unassign_workspaces_from_capacity: improved to accept in the Workspace list any mix of ID or Names
- revoke_external_data_share: accepts Workspace ID or Name
- scan_workspaces: accepts in the workspace list any mix of ID or Names
- scan_workspaces: accepts any number of workspaces in the list. Limit of 500 request for 1 hour and 100 workspaces per request stills applies. The limit is 50k workspaces.
- list_items: accepts workspace ID or Name
- list_items: accepts capacity ID or Name
- list_item_access_details: accepts item ID or Name
- list_item_access_details: accepts workspace ID or Name
- list_activity_events: added an option to return the original json, knowing the schema for event is not fix and too diverse, it helps for extraction
- _resolve_capacity_name_and_id: created to get a capacity ID and Name from the Capacity ID or Name
- _resolve_workspace_name_and_id: accepts ID or Name
- _resolve_item_id: accepts workspace ID or Name

### DIFF
--- a/src/sempy_labs/admin/_basic_functions.py
+++ b/src/sempy_labs/admin/_basic_functions.py
@@ -4,7 +4,6 @@ from uuid import UUID
 import sempy_labs._icons as icons
 from sempy.fabric.exceptions import FabricHTTPException
 from sempy_labs._helper_functions import (
-    resolve_workspace_name_and_id,
     pagination,
     _is_valid_uuid,
 )
@@ -12,6 +11,7 @@ import numpy as np
 import pandas as pd
 import time
 import urllib.parse
+from dateutil.parser import parse as dtparser
 from datetime import datetime
 
 
@@ -54,7 +54,7 @@ def list_workspaces(
             "The 'filter' parameter has been deprecated. Please remove this parameter from the function going forward."
         )
         del kwargs["filter"]
-
+    
     client = fabric.FabricRestClient()
 
     df = pd.DataFrame(
@@ -112,10 +112,11 @@ def list_workspaces(
             inplace=True,
         )
 
-        df["Capacity Id"] = df["Capacity Id"].str.lower()
+        df['Id'] = df['Id'].str.lower()
+        df['Capacity Id'] = df['Capacity Id'].str.lower()
 
         if workspace is not None and _is_valid_uuid(workspace):
-            df = df[df["Id"] == workspace]
+            df = df[df["Id"] == workspace.lower()]
 
     if skip is not None:
         df = df.tail(-skip)
@@ -125,84 +126,6 @@ def list_workspaces(
         df = df.head(top)
 
     return df
-
-
-def assign_workspaces_to_capacity(
-    source_capacity: Optional[str | UUID] = None,
-    target_capacity: Optional[str | UUID] = None,
-    workspace: Optional[str | List[str] | UUID | List[UUID]] = None,
-):
-    """
-    Assigns a workspace to a capacity. This function is the admin version.
-
-    This is a wrapper function for the following API: `Admin - Capacities AssignWorkspacesToCapacity <https://learn.microsoft.com/rest/api/power-bi/admin/capacities-assign-workspaces-to-capacity>`_.
-
-    Parameters
-    ----------
-    source_capacity : str | UUID, default=None
-        The name of the source capacity. If the Workspace is not specified, this is parameter mandatory.
-    target_capacity : str | UUID, default=None
-        The name of the target capacity.
-    workspace : str | List[str] | UUID | List[UUID], default=None
-        The name or id of the workspace(s).
-        Defaults to None which resolves to migrating all workspaces within the source capacity to the target capacity.
-    """
-    if target_capacity is None:
-        raise ValueError(
-            f"{icons.red_dot} The parameter 'target_capacity' is mandatory."
-        )
-
-    if source_capacity is None and workspace is None:
-        raise ValueError(
-            f"{icons.red_dot} The parameters 'source_capacity' or 'workspace' needs to be specified."
-        )
-
-    if workspace is None:
-        source_capacity_id = _resolve_capacity_name_and_id(source_capacity)[1]
-        dfW = list_workspaces(capacity=source_capacity_id)
-        workspaces = dfW["Id"].tolist()
-    else:
-        if isinstance(workspace, str):
-            workspace = [workspace]
-        if source_capacity is None:
-            dfW = list_workspaces()
-        else:
-            dfW = list_workspaces(capacity=source_capacity_id)
-        workspaces = dfW[dfW["Name"].isin(workspace)]["Id"].tolist()
-        workspaces = workspaces + dfW[dfW["Id"].isin(workspace)]["Id"].tolist()
-
-    if len(workspace) != len(workspaces):
-        raise ValueError(
-            f"{icons.red_dot} Some of the workspaces provided are not valid."
-        )
-
-    target_capacity_id = _resolve_capacity_name_and_id(target_capacity)[1]
-
-    workspaces = np.array(workspaces)
-    batch_size = 999
-    for i in range(0, len(workspaces), batch_size):
-        batch = workspaces[i : i + batch_size].tolist()
-        request_body = {
-            "capacityMigrationAssignments": [
-                {
-                    "targetCapacityObjectId": target_capacity_id.upper(),
-                    "workspacesToAssign": batch,
-                }
-            ]
-        }
-
-        client = fabric.FabricRestClient()
-
-        response = client.post(
-            "/v1.0/myorg/admin/capacities/AssignWorkspaces",
-            json=request_body,
-        )
-
-        if response.status_code != 200:
-            raise FabricHTTPException(response)
-    print(
-        f"{icons.green_dot} The workspaces have been assigned to the '{target_capacity}' capacity. A total of {len(workspaces)} were moved."
-    )
 
 
 def list_capacities(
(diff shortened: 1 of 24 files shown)


src/sempy_labs/admin/_basic_functions.py
```diff
@@ -15,87 +16,114 @@
...
+
+    if len(workspaces) > 0:
+        df = pd.DataFrame(workspaces)
+        df.rename(
+            columns={
+                "id": "Id",
+                "name": "Name",
+                "state": "State",
+                "type": "Type",
+                "capacityId": "Capacity Id",
```
m-kovalsky, line 106: Capacity Id must be in lowercase for consistency with semantic link functions.

src/sempy_labs/admin/_basic_functions.py
```diff
@@ -15,87 +16,114 @@
 
 
 def list_workspaces(
-    top: Optional[int] = 5000,
+    top: Optional[int] = None,
     filter: Optional[str] = None,
     skip: Optional[int] = None,
+    capacity: Optional[str] = None,
```
m-kovalsky, line 22: capacity & workspace parameters should be Optional[str | UUID] - correct?

src/sempy_labs/admin/_basic_functions.py
```diff
@@ -15,87 +16,114 @@
 
 
 def list_workspaces(
-    top: Optional[int] = 5000,
+    top: Optional[int] = None,
     filter: Optional[str] = None,
```
m-kovalsky, line 20: since the 'filter' parameter is deprecated, use **kwargs to tell the user so. See the example here: https://github.com/microsoft/semantic-link-labs/blob/main/src/sempy_labs/_model_bpa_rules.py

src/sempy_labs/admin/_basic_functions.py
```diff
@@ -105,32 +133,37 @@ def assign_workspaces_to_capacity(
 
     Parameters
     ----------
-    source_capacity : str
-        The name of the source capacity.
-    target_capacity : str
+    source_capacity : str, default=None
+        The name of the source capacity. If the Workspace is not specified, this is parameter mandatory.
```
m-kovalsky, line 137: the source_capacity & target_capacity should not be optional and should not default to None. These are mandatory.

src/sempy_labs/admin/_basic_functions.py
```diff
@@ -105,32 +133,37 @@ def assign_workspaces_to_capacity(
...
+    target_capacity : str, default=None
         The name of the target capacity.
     workspace : str | List[str], default=None
-        The name of the workspace(s).
+        The name or id of the workspace(s).
```
m-kovalsky, line 141: workspace : str | UUID | List[str] | List[UUID]. By the way, would this work if they put a list with a combination of names & ids?

src/sempy_labs/admin/_basic_functions.py
```diff
@@ -105,32 +133,37 @@ def assign_workspaces_to_capacity(
...
-    dfC_filt = dfC[dfC["Capacity Name"] == target_capacity]
-    target_capacity_id = dfC_filt["Capacity Id"].iloc[0]
+    if source_capacity is None and workspace is None:
+        raise ValueError(
+            "The parameters source_capacity or workspace needs to be specified."
+        )
 
     if workspace is None:
-        dfW = list_workspaces()
-        dfW = dfW[dfW["Capacity Id"].str.upper() == source_capacity_id.upper()]
```
m-kovalsky, line 129: when you update capacity Id in admin.list_workspaces to lowercase, make sure that the case matches here.

src/sempy_labs/admin/_basic_functions.py
ecotte: In this case is the rename, in the response we have capacityId and we are renaming it to "Capacity Id" like the rest of the pd.DataFarames.

src/sempy_labs/admin/_basic_functions.py
ecotte: we could do that, no problem. Just a consideration, it would help for documentation purpose, but to really pass a UUID you need to pass a parameter in UUID. 

That means that if you don´t pass it as "UUID(str({parameter}), version=4)" is going to be a str always.

But like I said, at the same time it helps to understand that it accepts an ID.

src/sempy_labs/admin/_basic_functions.py
ecotte: Like it. Will do it that way. :)

src/sempy_labs/admin/_basic_functions.py
ecotte: The source could be optional if we have the workspace list.

I created 2 validations about it in 144.

Now, I can change the order of the target_capacity and put it first, because that one is mandatory, and then just put the source_capacity and workspace as optional and leave the validation that one or the other needs to be there.

It helps in scenarios that I had with a client that wanted to move WS from multiple capacities to a single one, or just a group a WS to a Capacity.

src/sempy_labs/admin/_basic_functions.py
ecotte: Same as before. We can use it for documentation purpose, but in case of a mix of Names and IDs will always be a List[str] and it would be only a UUID or List[UUID] if you specified before passing the parameter.

src/sempy_labs/admin/_basic_functions.py
ecotte: I can change it to be lower in both cases and match the rest of the code. In this case it works anyway because it´s matching been upper, but I agree is not consistent with the rest.

src/sempy_labs/admin/_basic_functions.py
m-kovalsky: the value needs to be lowercase, not the name

ecotte: Values are returned in lowercase now.

src/sempy_labs/admin/_basic_functions.py
```diff
@@ -112,10 +112,11 @@ def list_workspaces(
             inplace=True,
         )
 
-        df["Capacity Id"] = df["Capacity Id"].str.lower()
+        df['Id'] = df['Id'].str.lower()
```
m-kovalsky: no need for workspace Id -> lower

src/sempy_labs/admin/_basic_functions.py
```diff
@@ -624,12 +927,12 @@ def list_item_access_details(
 
     Parameters
     ----------
-    item_name : str
-        Name of the Fabric item.
-    type : str
-        Type of Fabric item.
+    item : str
+        Name or id of the Fabric item.
+    type : str, default=None
```
m-kovalsky, line 932: type is necessary. need to raise an error if this is not provided.

src/sempy_labs/admin/_basic_functions.py
```diff
@@ -4,99 +4,186 @@
...
+        df["Capacity Id"] = df["Capacity Id"].str.lower()
+
+        if workspace is not None and _is_valid_uuid(workspace):
+            df = df[df["Id"] == workspace.lower()]
+
     if skip is not None:
-        url = f"{url}&$skip={skip}"
-    if filter is not None:
-        url = f"{url}&$filter={filter}"
+        df = df.tail(-skip)
```
eisber, line 122: there's no filtering equivalent on the new API? if so, I'd suggest to remove skip and filter as these suggest that you can somehow improve performance of the call, rather everything is fetched anyway.

src/sempy_labs/admin/_basic_functions.py
```diff
@@ -105,32 +192,44 @@ def assign_workspaces_to_capacity(
...
+        if source_capacity is None:
+            dfW = list_workspaces()
+        else:
+            dfW = list_workspaces(capacity=source_capacity_id)
         workspaces = dfW[dfW["Name"].isin(workspace)]["Id"].tolist()
+        workspaces = workspaces + dfW[dfW["Id"].isin(workspace)]["Id"].tolist()
+
+    if len(workspace) != len(workspaces):
+        raise ValueError(
+            f"{icons.red_dot} Some of the workspaces provided are not valid."
```
eisber, line 229: can you output which ones are not valid?

src/sempy_labs/admin/_basic_functions.py
```diff
@@ -563,7 +729,24 @@ def list_datasets() -> pd.DataFrame:
...
+    if top is not None:
+        params["$top"] = top
+
+    if filter is not None:
+        params["$filter"] = filter
+
+    if skip is not None:
+        params["$skip"] = skip
+
+    url_parts = list(urllib.parse.urlparse(url))
```
eisber, line 745: move to utility function. e.g. url = build_url(url, params)

eisber (changes requested):

src/sempy_labs/admin/_basic_functions.py
ecotte: added in the parameters validation and it raises the error.

src/sempy_labs/admin/_basic_functions.py
ecotte: Added the list of invalid workspaces.

src/sempy_labs/admin/_basic_functions.py
ecotte: Added the function _build_url to _helper_functions. And refactoring the function codes.

eisber (approved):

PR #298, opened 2024-11-24
M kovalsky/alttext title subtitle

m-kovalsky:


src/sempy_labs/report/_report_helper.py
```diff
@@ -252,3 +252,59 @@ def find_entity_property_pairs(data, result=None, keys_path=None):
...
+
+def _get_agg_type_mapping() -> dict:
+
+    schema_url = "https://developer.microsoft.com/json-schemas/fabric/item/report/definition/semanticQuery/1.2.0/schema.json"
+    response = requests.get(schema_url)
+    schema = response.json()
+    aggtypes_schema = schema.get("definitions", {}).get("QueryAggregateFunction", {})
+
+    agg_type_map = {}
+    agg_type_map = {
```
eisber, line 265: can you add an example of what you are extracting

src/sempy_labs/report/_report_helper.py
```diff
@@ -252,3 +252,59 @@ def find_entity_property_pairs(data, result=None, keys_path=None):
...
+            expr_json.get("Aggregation", {})
+            .get("Expression", {})
+            .get("Column", {})
+            .get("Property", "Column not found")
+        )
+        function_id = expr_json.get("Aggregation", {}).get("Function", "-1")
+        function = agg_type_map.get(function_id)
+        expr = f"{function}('{entity}'[{column}])"
+    elif expr_type == "Measure":
+        entity = (
```
eisber, line 299: what's the use of entity here?

src/sempy_labs/report/_report_helper.py
```diff
@@ -252,3 +252,59 @@ def find_entity_property_pairs(data, result=None, keys_path=None):
...
+            .get("SourceRef", {})
+            .get("Entity", "Entity not found")
+        )
+        column = (
+            expr_json.get("Aggregation", {})
+            .get("Expression", {})
+            .get("Column", {})
+            .get("Property", "Column not found")
+        )
+        function_id = expr_json.get("Aggregation", {}).get("Function", "-1")
```
eisber, line 295: what's the expectation if neither entity and/or column doesn't resolve? throw an exception? just put a string into the outcome? if users want to filter, they'd have to search for "Column not found"

eisber (changes requested):

src/sempy_labs/report/_report_helper.py
m-kovalsky: added

src/sempy_labs/report/_report_helper.py
m-kovalsky: removed

src/sempy_labs/report/_report_helper.py
m-kovalsky: updated these to unknown

eisber (approved):

Issue #300, opened 2024-11-25
Plans for exporting/importing TMDL for semantic models?


jamescl604:
Just curious if there are plans to support the ability to export a semantic model to TMDL as well as support imports from TMDL.   I'd like to crawl all the model definitions as text and use GPT to do certain tasks, like documentation to translations.

m-kovalsky:
Yes. In what format would you prefer the output to be in. With TMSL it's just a single .bim file so it's easy to handle. With TMDL it's a whole bunch of files so the output is more complicated.

jamescl604:
I was looking at this type of syntax which is available from AMO:  https://learn.microsoft.com/en-us/analysis-services/tmdl/tmdl-how-to | https://learn.microsoft.com/en-us/dotnet/api/microsoft.analysisservices.tmdlserializer?view=analysisservices-dotnet

It supports use cases of serializing to a folder, to a text variable, and the reverse which is pretty handy if it was only available from Python : )

m-kovalsky:
Added the get_semantic_model_definition function in [0.8.9](https://github.com/microsoft/semantic-link-labs/releases/tag/0.8.9).

PR #308, opened 2024-11-27
added get_semantic_model_definition

m-kovalsky:
#300

### DIFF
--- a/src/sempy_labs/_generate_semantic_model.py
+++ b/src/sempy_labs/_generate_semantic_model.py
@@ -2,7 +2,7 @@ import sempy.fabric as fabric
 import pandas as pd
 import json
 import os
-from typing import Optional
+from typing import Optional, List
 from sempy_labs._helper_functions import (
     resolve_lakehouse_name,
     resolve_workspace_name_and_id,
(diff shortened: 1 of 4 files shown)


src/sempy_labs/_generate_semantic_model.py
```diff
@@ -384,12 +369,83 @@ def get_semantic_model_bim(
...
+    pandas.DataFrame | dict
+        A pandas dataframe with the semantic model definition or the file or files comprising the semantic model definition.
+
+    """
+
+    valid_formats = ['TMSL', 'TMDL']
+
+    format = format.upper()
+    if format == 'BIM':
+        format = "TMSL"
```
eisber, line 414: without a warning this seems strange? or add it to the docs?

src/sempy_labs/_generate_semantic_model.py
```diff
@@ -384,12 +369,83 @@ def get_semantic_model_bim(
...
+    files = result["definition"]["parts"]
+
+    if return_dataframe:
+        return pd.json_normalize(files)
+    elif format == 'TMSL':
+        payload = next(
+            (part["payload"] for part in files if part["path"] == "model.bim"),
+            None
+        )
+        return json.loads(_decode_b64(payload))
```
eisber, line 436: throw proper exception if model.bim not found?

src/sempy_labs/_generate_semantic_model.py
m-kovalsky: tmsl is the format of a bim file. I updated the function documentation to clarify this.

src/sempy_labs/_generate_semantic_model.py
m-kovalsky: Every model must have a bim file behind it. Semantic models are naturally in .bim (TMSL) format. This API allows for returning the model in TMDL format as well.

eisber (approved):

PR #315, opened 2024-11-29
edcotte/assign_workspace

ecotte:
Fix on assign workspace with some combinations. Tested scenarios:

- No parameters
- No source or workspace list
- Only target capacity and a workspace list
- All 3 parameters assigned. Will look for the WS in the source_capacity
- Target Capacity and WS as Text
- Target capacity and list of WSs
- Only source_capacity and target_capacity, WS as None
- Only target_capacity in ID and WS in ID
- Only source_capacity in ID and target_capacity in ID


m-kovalsky (approved):
Thanks for the fix

PR #319, opened 2024-11-29
Update _data_pipelines.py

hurtn:
Added update data pipeline definition method

### DIFF
--- a/src/sempy_labs/_data_pipelines.py
+++ b/src/sempy_labs/_data_pipelines.py
@@ -170,3 +170,53 @@ def get_data_pipeline_definition(
     payload = content["payload"].iloc[0]
 
     return _decode_b64(payload)
+
+
+def update_data_pipeline_definition(
+    name: str, pipeline_content: str, workspace: Optional[str] = None
+):
+    """
+    Updates an existing data pipeline with a new definition.
+
+    Parameters
+    ----------
+    name : str
+        The name of the data pipeline.
+    pipeline_content : str
+        The data pipeline content (not in Base64 format).
+    workspace : str, default=None
+        The name of the workspace.
+        Defaults to None which resolves to the workspace of the attached lakehouse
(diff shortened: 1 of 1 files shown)


src/sempy_labs/_data_pipelines.py
```diff
@@ -170,3 +170,53 @@ def get_data_pipeline_definition(
     payload = content["payload"].iloc[0]
 
     return _decode_b64(payload)
+
+
+def update_data_pipeline_definition(
+    name: str, pipeline_content: str, workspace: Optional[str] = None
```
m-kovalsky, line 176: won't the content be in dict format? Check out the _conv_b64 function in sempy_labs._helper_functions.

let's use the parameter' content' instead of 'pipeline_content'.

src/sempy_labs/_data_pipelines.py
```diff
@@ -170,3 +170,53 @@ def get_data_pipeline_definition(
     payload = content["payload"].iloc[0]
 
     return _decode_b64(payload)
+
+
+def update_data_pipeline_definition(
```
m-kovalsky, line 175: don't forget to add this function to the __init__.py file. otherwise it won't show up for users.

src/sempy_labs/_data_pipelines.py
hurtn: Sure, I can change it to dict, was using your update notebook definition as a template https://github.com/microsoft/semantic-link-labs/blob/ebd7b84ec7b059b8d46c1bcd91092fddf55fbf7a/src/sempy_labs/_notebooks.py#L180 
LMK

m-kovalsky: ah but the content/definition for a notebook isn't in json format. it's in .py format.

hurtn: gotcha, make sense. i'll update tomorrow!

PR #320, opened 2024-11-29
Update _git.py

hurtn:
Fix LRO bug - no result is returned from a commit to git operation therefore added True boolean param to use status code instead. Added check to see whether any changes exist before calling commit operation and added message if no modified items found. Fixed small typo.

### DIFF
--- a/src/sempy_labs/_git.py
+++ b/src/sempy_labs/_git.py
@@ -277,44 +277,50 @@ def commit_to_git(
     workspace, workspace_id = resolve_workspace_name_and_id(workspace)
 
     gs = get_git_status(workspace=workspace)
-    workspace_head = gs["Workspace Head"].iloc[0]
+    if len(gs.index) > 0:
+        workspace_head = gs["Workspace Head"].iloc[0]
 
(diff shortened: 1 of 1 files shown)


src/sempy_labs/_git.py
```diff
@@ -277,46 +277,52 @@ def commit_to_git(
     workspace, workspace_id = resolve_workspace_name_and_id(workspace)
 
     gs = get_git_status(workspace=workspace)
-    workspace_head = gs["Workspace Head"].iloc[0]
+    if len(gs.index) > 0:
+        workspace_head = gs["Workspace Head"].iloc[0]
```
m-kovalsky, line 281: instead should have a raise valuerror. something like this.

if len(gs) == 0:
raise ValueError(f"{icons.red_dot}..."
workspace_head = gs["Workspace Head"].ilocl[0]

hurtn: Perhaps, I would suggest however it's not an error as such, if anything a notification/warning

m-kovalsky (approved):
Thanks Nick!

Issue #317, opened 2024-11-29
admin.list_activity_events stopped working for pandas.DataFrame.


Jai-Prakash-HU:
after this recent update admin.list_activity_events  doesn't return output for below code. 
It keeps running for forever I checked until 38 minutes, it did not return anything. 
It stopped working in recent updated. 

![Image](https://github.com/user-attachments/assets/a45d08f6-0bf8-4ac2-85dd-34e7e50362d5)

Also, If I change the return_dataframe = False  it gives output in few seconds in Json format. But again, now it is not considering the parameter. You can check below it is giving me all the types of events but I have applied activity_filter =  'EditDataset'. I am not sure if it is considering the user filter. Same issue happened last month but it was fixed. 

![Image](https://github.com/user-attachments/assets/500cf479-1a6d-4953-bd14-7ac1af1738f1)

I had automated the job using this function with pandas.DataFrame result getting captured and saved in table. But now jobs are running from last 8 hours. No result. Could you please fix the pandas.DataFrame output. 
also please maintain the number of columns as output as I am appending the result in existing tables. So, if the output fields are stable over the time, then it would be easier for me to maintain it for log purpose. 

Recently I faced a difference in the output columns after recent update of function admin.list_workspaces. Now Is_On_Dedicated_Capacity  is missing. I was using Is_On_Dedicated_Capacity = 1  OfCourse I can adjust my job but please maintain the number of fields in pre and post update of function.

m-kovalsky:
Fixed in [0.8.9](https://github.com/microsoft/semantic-link-labs/releases/tag/0.8.9).

PR #323, opened 2024-12-01
fixed issue 317

m-kovalsky:
#317

### DIFF
--- a/src/sempy_labs/admin/_basic_functions.py
+++ b/src/sempy_labs/admin/_basic_functions.py
@@ -782,34 +782,18 @@ def list_activity_events(
     )
 
     resposeJson = {"activityEventEntities": []}
-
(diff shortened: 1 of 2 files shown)


src/sempy_labs/admin/_basic_functions.py
```diff
@@ -853,11 +837,9 @@ def list_activity_events(
 
     if return_dataframe:
         df["Creation Time"] = pd.to_datetime(df["Creation Time"])
-        activity_events = df
+        return df
     else:
-        activity_events = resposeJson
-
-    return activity_events
+        return resposeJson
```
eisber, line 842: don't have to fix here. python style guide: variables = snake case, class names = pascal case

eisber (approved):

PR #325, opened 2024-12-02
add token provider to admin.list_workspaces

m-kovalsky:


src/sempy_labs/admin/_basic_functions.py
```diff
@@ -694,10 +718,11 @@ def list_workspace_access_details(
             "Workspace Name",
             "Workspace Id",
             "Workspace Role",
+            s,
```
ecotte, line 721: There´s a typo in this line, it shouldn´t have the "s," column

src/sempy_labs/admin/_basic_functions.py
```diff
@@ -545,7 +562,7 @@ def list_datasets(
         ]
     )
 
-    client = fabric.PowerBIRestClient()
+    client = fabric.PowerBIRestClient(token_provider=token_provider)
```
ecotte, line 565: If we use the PowerBIRestClient the token_provider doesn´t work, but with the FabricRestClient it does work.

src/sempy_labs/admin/_basic_functions.py
```diff
@@ -782,7 +810,7 @@ def list_activity_events(
     )
 
     response_json = {"activityEventEntities": []}
-    client = fabric.PowerBIRestClient()
+    client = fabric.PowerBIRestClient(token_provider=token_provider)
```
ecotte, line 813: If we use the PowerBIRestClient the token_provider doesn´t work, but with the FabricRestClient it does work.

src/sempy_labs/admin/_basic_functions.py
```diff
@@ -955,7 +999,7 @@ def list_reports(
 
     url.rstrip("$").rstrip("?")
 
-    client = fabric.PowerBIRestClient()
+    client = fabric.PowerBIRestClient(token_provider=token_provider)
```
ecotte, line 1002: If we use the PowerBIRestClient the token_provider doesn´t work, but with the FabricRestClient it does work.

PR #329, opened 2024-12-04
added capacity assignment status

m-kovalsky:


src/sempy_labs/admin/_basic_functions.py
```diff
@@ -901,14 +900,20 @@ def _resolve_workspace_name_and_id(
...
-        raise ValueError(f"{icons.red_dot} The '{workspace}' workspace was not found.")
+    if workspace is None:
+        workspace_id = fabric.get_workspace_id()
+        workspace_name = fabric.resolve_workspace_name(workspace_id)
+    else:
+        dfW = list_workspaces(workspace=workspace)
+        try:
+            workspace_name = dfW["Name"].iloc[0]
+            workspace_id = dfW["Id"].iloc[0]
+        except Exception:
```
eisber, line 911: can you check on dfW empty instead of catch exception?

eisber (changes requested):

eisber (approved):

PR #340, opened 2024-12-08
Added 17 missing columns to the list_activity_events function

IC-XC-NIKA94:
There are 17 additional columns that are not currently in the `list_activity_events` function. I've added them in this PR, and confirmed on my end that all of them are pulling non-null values in. Some of these are necessary to see, for example, how a PBI report is being consumed (by App or Workspace).


IC-XC-NIKA94:

IC-XC-NIKA94:
Here are the list of new columns being returned:

 "Is Success",
            "Activity Id",
            "Item Name",
            "Dataset Name",
            "Report Name",
            "Capacity Id",
            "Capacity Name",
            "App Name",
            "Dataset Id",
            "Report Id",
            "Artifact Id",
            "Artifact Name",
            "Report Type",
            "App Report Id",
            "Distribution Method",
            "Consumption Method",
            "Artifact Kind",

m-kovalsky (approved):

PR #354, opened 2024-12-15
M kovalsky/authentication azure

m-kovalsky:


src/sempy_labs/_capacities.py
```diff
@@ -143,29 +136,42 @@ def create_fabric_capacity(
...
-        key_vault_uri=key_vault_uri,
-        key_vault_tenant_id=key_vault_tenant_id,
-        key_vault_client_id=key_vault_client_id,
-        key_vault_client_secret=key_vault_client_secret,
-    )
-
-    resource_client = ResourceManagementClient(credential, azure_subscription_id)
+    if token_provider is None:
+        token_provider = ServicePrincipalTokenProvider.from_azure_key_vault(
+            key_vault_uri=kwargs["key_vault_uri"],
```
eisber, line 141: I'm confused, you're removing the args, but then fetching the out of kwargs?

seems messy to support this API? can't you just break it and version bump?

src/sempy_labs/_capacities.py
```diff
@@ -634,57 +611,665 @@ def check_fabric_capacity_name_availablility(
...
+    df = pd.DataFrame(
+        columns=[
+            "Storage Account Id",
+            "Storage Account Name",
+            "Kind",
+            "Location",
+            "Sku Name",
+            "Sku Tier",
+            "Is HNS Enabled",
+            "Creation Time",
```
eisber, line 1081: convert time to proper date time?

eisber (changes requested):

src/sempy_labs/_capacities.py
m-kovalsky: i don't want to break this one as it is in use in major scenarios. will remove this part in the future (in a month or so).

eisber (approved):

PR #358, opened 2024-12-16
added get_tenant_id

m-kovalsky:


src/sempy_labs/_helper_functions.py
```diff
@@ -1371,3 +1377,15 @@ def _is_valid_uuid(
         return True
     except ValueError:
         return False
+
+
+def _get_x_id(name: str):
```
eisber, line 1382: rename to get_fabric_context_setting  ? get_x_id seems a bit mysterious

eisber (approved):

PR #372, opened 2024-12-19
fixed dependencies and unqualified_columns

m-kovalsky:


src/sempy_labs/_model_dependencies.py
```diff
@@ -189,12 +190,24 @@ def get_model_calc_dependencies(
...
+
     # Initialize dependency DataFrame with 'Done' status
     df = dep.copy()
     objs = {"Measure", "Calc Column", "Calculation Item", "Calc Table"}
     df["Done"] = (
         df["Referenced Object Type"].apply(lambda x: x not in objs).astype(bool)
     )
+
+    # Set to track visited dependencies to prevent circular references
+    visited = set(
```
eisber, line 202: is this the same as df[["Full Object Name", "Referenced Full Object Name",...]].drop_duplicates()

eisber (approved):

src/sempy_labs/_model_dependencies.py
m-kovalsky: set creates a tuple, drop_duplicates creates a dataframe.

PR #386, opened 2024-12-30
added optimize delta tables to support all delta tables

m-kovalsky:


src/sempy_labs/_optimize.py
```diff
@@ -0,0 +1,80 @@
...
+        delta_tables = pd.DataFrame(data)
+
+    spark = SparkSession.builder.getOrCreate()
+
+    for _, r in (bar := tqdm(delta_tables.iterrows())):
+        tableName = r["Table Name"]
+        tablePath = r["Location"]
+        bar.set_description(f"Optimizing the '{tableName}' table...")
+        deltaTable = DeltaTable.forPath(spark, tablePath)
+        deltaTable.optimize().executeCompaction()
```
eisber, line 80: why would you use the notebook instead of this? https://learn.microsoft.com/en-us/rest/api/fabric/lakehouse/background-jobs/run-on-demand-table-maintenance?tabs=HTTP

eisber (changes requested):

src/sempy_labs/_optimize.py
m-kovalsky: removed this function for now

src/sempy_labs/_helper_functions.py
```diff
@@ -52,6 +52,24 @@ def create_abfss_path(
...
+    default_file_storage = _get_fabric_context_setting(name="fs.defaultFS")
+
+    return default_file_storage.split("@")[-1][:-1]
+
+
+def _split_abfss_path(path: str) -> Tuple[UUID, UUID, str]:
+
+    # Extracts the workspace ID, item ID and delta table name from the abfss path.
+
+    workspace_id = path.split("abfss://")[1].split("@")[0]
```
eisber, line 66: in general I'd recommend to https://docs.python.org/3/library/urllib.parse.html

src/sempy_labs/_optimize.py
eisber: function is still there?

src/sempy_labs/_helper_functions.py
m-kovalsky: fixed

src/sempy_labs/_optimize.py
m-kovalsky: removed

eisber (approved):

PR #392, opened 2025-01-03
Fix model bpa rules

bozhou8:
Fix "Check if bi-directional and many-to-many relationships are valid" and "Set IsAvailableInMdx to true on necessary columns" model BPA rules:

1. "Check if bi-directional and many-to-many relationships are valid": fix missing comma.
2. "Set IsAvailableInMdx to true on necessary columns": fix expression determining if generator is empty


m-kovalsky (approved):
Thanks @bo-zhou-ms!

PR #395, opened 2025-01-06
added delta analyzer

m-kovalsky:


src/sempy_labs/_dax.py
```diff
@@ -257,3 +260,218 @@ def get_dax_query_memory_size(
...
+            for name, dax in dax_queries.items():
+
+                if clear_cache_before_run:
+                    clear_cache(dataset=dataset, workspace=workspace)
+                if refresh_type is not None:
+                    refresh_semantic_model(
+                        dataset=dataset, workspace=workspace, refresh_type=refresh_type
+                    )
+
+                fabric.evaluate_dax(
```
eisber, line 381: what's the difference between EVALUATE {1} and just {1}. add a comment?

src/sempy_labs/_delta_analyzer.py
```diff
@@ -0,0 +1,272 @@
...
+        A dictionary of pandas dataframes showing semantic model objects which violated the best practice analyzer rules.
+    """
+
+    prefix = "SLL_DeltaAnalyzer_"
+    now = datetime.datetime.now()
+    workspace_id = fabric.get_workspace_id()
+    workspace_name = fabric.resolve_workspace_name(workspace=workspace_id)
+    lakehouse_id = fabric.get_lakehouse_id()
+    lakehouse_name = resolve_lakehouse_name()
+    default_file_storage = _get_fabric_context_setting(name="fs.defaultFS")
```
eisber, line 46: can you figure from this if a lakehouse is attached? if so, throw an error?

eisber (approved):

src/sempy_labs/_dax.py
m-kovalsky: i only use evaluate {1}. this is to initiate cache warming

src/sempy_labs/_delta_analyzer.py
m-kovalsky: done

PR #397, opened 2025-01-07
Add notebook to list large semantic models

andrewluotechnologies:
When migrating capacities, extra action is needed when moving workspaces between capacities in different regions if large semantic models exist in the workspace. It would be useful to have a notebook to list those semantic models.

### DIFF
--- a/notebooks/List Large Semantic Models.ipynb
+++ b/notebooks/List Large Semantic Models.ipynb
@@ -1,0 +1,116 @@
+{
+    "cells": [
+        {
+            "cell_type": "code",
+            "source": [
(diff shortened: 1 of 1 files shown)


m-kovalsky:
I don't really think this warrants its own notebook. It can be done in 2 lines of code. Maybe in a script repository would be a better idea.

JeridInsight:
I do this when I run my daily inventory notebook (Get All Datasets): 

``#Get a list of all the datasets in the entire tenant
dfDatasets = spark.createDataFrame(labs.admin.list_datasets())
display(dfDatasets)

PR #400, opened 2025-01-09
added sp context manager

m-kovalsky:


src/sempy_labs/_authentication.py
```diff
@@ -156,3 +161,37 @@ def _get_headers(
...
+        Azure Key Vault URI.
+    key_vault_tenant_id : str
+        Name of the secret in the Key Vault with the Fabric Tenant ID.
+    key_vault_client_id : str
+        Name of the secret in the Key Vault with the Service Principal Client ID.
+    key_vault_client_secret : str
+        Name of the secret in the Key Vault with the Service Principal Client Secret.
+    """
+    try:
+        auth.token_provider = ServicePrincipalTokenProvider.from_azure_key_vault(
```
whiskyboy, line 189: We should save the prior state of the auth.token_provider and restore it in the finally block

src/sempy_labs/_auth.py
```diff
@@ -0,0 +1 @@
+token_provider = None
```
whiskyboy, line 1: Consider make it a context var for thread-safety?

src/sempy_labs/admin/_domains.py
```diff
@@ -68,7 +60,7 @@ def list_domains(
         ]
     )
 
-    client = fabric.FabricRestClient(token_provider=token_provider)
+    client = fabric.FabricRestClient(token_provider=auth.token_provider)
```
whiskyboy, line 63: Shouldn't it be wrapped with service_principal_authentication() context?

m-kovalsky: what do you mean by this?

src/sempy_labs/_auth.py
m-kovalsky: made a change for this

src/sempy_labs/_authentication.py
m-kovalsky: made a change for this

src/sempy_labs/_authentication.py
```diff
@@ -156,3 +161,51 @@ def _get_headers(
...
+            key_vault_tenant_id=key_vault_tenant_id,
+            key_vault_client_id=key_vault_client_id,
+            key_vault_client_secret=key_vault_client_secret,
+        )
+    )
+    try:
+        yield
+    finally:
+        # Restore the prior state
+        if prior_token is None:
```
eisber, line 208: why do you need to check if it's None, if you set it to None anyway?

src/sempy_labs/_capacities.py
```diff
@@ -361,10 +351,9 @@ def delete_embedded_capacity(
         The Azure subscription ID.
     resource_group : str
         The name of the Azure resource group.
-    token_provider : TokenProvider, default=None
-        The token provider for authentication, created by using the ServicePrincipalTokenProvider class.
     """
 
+    token_provider = auth.token_provider.get()
     if token_provider is None:
```
eisber, line 357: wouldn't it be simpler and cleaner if we just throw if no context provider is set?
which can be done generically in token_provider.get()

m-kovalsky: this code snippet ensures continuity for folks who were using the parameters which were in the function before implementing the SP context manager.

src/sempy_labs/_authentication.py
```diff
@@ -156,3 +161,51 @@ def _get_headers(
         headers["Content-Type"] = "application/json"
 
     return headers
+
+
+token_provider = contextvars.ContextVar("token_provider", default=None)
```
bozhou8, line 166: Seems need to make `default=SynapseTokenProvider()`? Otherwise calling out of the SP context with fill fail.

m-kovalsky: but doesn't None default to the synapsetokenprovider()?

src/sempy_labs/admin/_domains.py
whiskyboy: Oh pls ignore my previous comment. Just noticed you have it wrapped with service_principal_authentication() context in the ipynb file, which is fine.

A small suggestion: since `auth.token_provider` is a `contextvar`, we should call the `get()` method to retrieve its value

m-kovalsky: i updated all instances of auth.token_provider to auth.token_provider.get()

whiskyboy: Sorry to not check your latest code... We can close this comment now :)

src/sempy_labs/admin/_basic_functions.py
```diff
@@ -524,7 +505,6 @@ def list_datasets(
     top: Optional[int] = None,
     filter: Optional[str] = None,
     skip: Optional[int] = None,
-    token_provider: Optional[TokenProvider] = None,
```
whiskyboy, line 527: Is it OK to remove the parameter from the function signaure? Will it break cx existing codes if they are already using the `token_provider` paramter? Or is this breaking is by-design?

whiskyboy: Ah I see. The design LGTM now!

eisber (approved):

PR #404, opened 2025-01-12
Adds a new function that allows users to query kql databases

DAXNoobJustin:


DAXNoobJustin:

PR #405, opened 2025-01-12
Add query kql database function

DAXNoobJustin:
Adds a function that allows the user to query a kql database


m-kovalsky:
Thanks. I already have a [PR](https://github.com/microsoft/semantic-link-labs/pull/297) to do the same type of thing (plus more).

Issue #423, opened 2025-01-29
sempy_labs.admin.list_tenant_settings() returns empty df


TheoPatt:
Hi sempy-labs team,

**Describe the bug**
since the new release of version 0.9.1 the function sempy_labs.admin.list_tenant_settings() returns an empty df. 
I tested this in two different tenants. It used to work just fine around two weeks ago. I am still the admin of both tenants.

**To Reproduce**
update environment to semantic-link-labs 0.9.1
execute notebook with the following code:

import sempy_labs as labs
from sempy_labs import admin

df = labs.admin.list_tenant_settings()

display(df)

**Expected behavior**
list of tenant settings

**Screenshots**
see attachment

**Desktop (please complete the following information):**
-browser: firefox

Thank you for any piece of information which could help me identify the underlying problem :-)!

m-kovalsky:
The API now returns 'value' instead of 'tenantSettings' which is not updated in the documentation. @hurtn is making a PR to resolve this issue. Thanks for reporting.

TheoPatt:
Thanks for the quick feedback and resolution!

m-kovalsky:
This actually should be working already. The product team made a hotfix which returns both 'tenantSettings' and 'value'.

PR #429, opened 2025-01-30
Fix issue #423

hurtn:
Fix issue where list_tenant_settings API returns an unexpected response key.

### DIFF
--- a/src/sempy_labs/admin/_basic_functions.py
+++ b/src/sempy_labs/admin/_basic_functions.py
@@ -329,7 +329,7 @@ def list_tenant_settings() -> pd.DataFrame:
 
     response = _base_api(request="/v1/admin/tenantsettings", client="fabric_sp")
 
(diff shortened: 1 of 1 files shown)


src/sempy_labs/admin/_basic_functions.py
```diff
@@ -329,7 +329,12 @@ def list_tenant_settings() -> pd.DataFrame:
 
     response = _base_api(request="/v1/admin/tenantsettings", client="fabric_sp")
 
-    for i in response.json().get("tenantSettings", []):
+    if 'settingName' in response.json():
```
m-kovalsky, line 332: now i just get 'value'. I think we can change this to:

for i in response.json().get("value", []):

hurtn: Done

m-kovalsky (approved):
Thank you!

PR #431, opened 2025-01-31
Completed all wrapper functions for item scheduling

nielsvdc:
Completed all the wrapper functions for scheduling Fabric items. Refactored and tested the existing functions, because they were working.

I noticed that we need some consistency in the function argument naming. For job scheduling specifying the item type is required when using the item name instead of the item id. I renamed all type arguments to item_type, as this is also the most used name in other functions throughout the project.


nielsvdc:
@m-kovalsky Could you have a look at this? I'm hoping these changes would be available with the next release.

m-kovalsky:
Thanks. Please do not change parameter names as it will break existing functionality.

nielsvdc:
@m-kovalsky I fixed to parameter name of type to the original name for all functions

PR #439, opened 2025-02-03
Fix Issue 423 take 2

hurtn:
Adding conditional logic in list_tenant_settings to cater for different API response keys


m-kovalsky:
Thanks Nick. I don't think we need this for now as the previous PR is working by getting the 'value'. Will close this for now.

PR #469, opened 2025-02-13
Add DAX Performance Testing notebook and new get capacity status func…

DAXNoobJustin:
This PR introduces a new Jupyter Notebook—DAX Performance Testing—designed to help users test DAX performance on their Semantic Models using Semantic Link (Labs) functionality. The notebook performs the following functions:

Execute DAX queries from an Excel file against Power BI/Fabric models.
Measure query performance under various cache states (cold, warm, and hot).
Manage capacity operations (pause/resume) to simulate different caching scenarios.
Log and persist performance metrics to a Lakehouse table.


DAXNoobJustin:
@m-kovalsky 

I added a new function that returns the status of a target capacity. I noticed that you have another function, update_fabric_capacity, that includes the same call and extracts other properties from the provided dictionary. Should I leave my function as is or create a more generic function that returns all the metadata provided from that call and refactor my new function and yours?

m-kovalsky:
We already have a function for this: https://semantic-link-labs.readthedocs.io/en/stable/sempy_labs.admin.html#sempy_labs.admin.get_capacity_state

DAXNoobJustin:
Oh great! I'm not sure how I missed that one. I will update my notebook.

m-kovalsky:
Hi Justin - thanks for the notebook however it is much too complicated for many users. The main purpose of semantic link labs is to abstract the technical challenges so that the user is left with something intuitive - even for someone who doesn't really know python.

DAXNoobJustin:
Makes sense! Thanks for the feedback.

PR #490, opened 2025-02-18
Update _git.py - added param to initialize_git_connection fn

hurtn:
Added initializationStrategy strategy parameter. Used when content exists in the workspace prior to calling initialize connection

### DIFF
--- a/src/sempy_labs/_git.py
+++ b/src/sempy_labs/_git.py
@@ -264,7 +264,7 @@ def get_git_connection(workspace: Optional[str | UUID] = None) -> pd.DataFrame:
     return df
 
 
-def initialize_git_connection(workspace: Optional[str | UUID] = None) -> str:
+def initialize_git_connection(workspace: Optional[str | UUID] = None, initialization_strategy: Optional[str] = 'None') -> str:
     """
     Initializes a connection for a workspace that is connected to Git.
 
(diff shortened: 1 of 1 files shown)


src/sempy_labs/_git.py
```diff
@@ -276,6 +276,11 @@ def initialize_git_connection(workspace: Optional[str | UUID] = None) -> str:
         The Fabric workspace name or ID.
         Defaults to None which resolves to the workspace of the attached lakehouse
         or if no lakehouse attached, resolves to the workspace of the notebook.
+    initialization_strategy : str
```
m-kovalsky, line 279: initialization_strategy : str, default='None'

src/sempy_labs/_git.py
```diff
@@ -285,8 +290,14 @@ def initialize_git_connection(workspace: Optional[str | UUID] = None) -> str:
 
     (workspace_name, workspace_id) = resolve_workspace_name_and_id(workspace)
```
m-kovalsky, line 292: add a check that the initialization_strategy is one of the 3 valid options. raise ValueError if it's not in those 3 options. This is just to make better error messages.

m-kovalsky (changes requested):
Thanks Nick!

src/sempy_labs/_git.py
a-holm: Just a helpful note for @hurtn: You can look at how `conflict_resolution_policy` is validated in the `update_from_git` function within the same file. If you want to do it consistent in the same style and so on.

PR #493, opened 2025-02-19
M kovalsky/enhancedeploysemanticmodel

m-kovalsky:


src/sempy_labs/tom/_model.py
```diff
@@ -4677,6 +4699,183 @@ def remove_role_member(self, role_name: str, member: str | List[str]):
...
+            .get("create")
+            .get("database")
+        )
+
+    def _reduce_model(self, perspective_name: str):
+        """
+        Reduces a model's objects based on a perspective. Adds the dependent objects within a perspective to that perspective.
+        """
+
+        from sempy_labs._model_dependencies import get_model_calc_dependencies
```
eisber, line 4725: is there a more generic use-case in the sense of reducing a semantic model to what is needed based on specific elements? (e.g. perspective, report, ...)?

eisber (approved):

src/sempy_labs/tom/_model.py
m-kovalsky: perspective is the simplest. based on a report is also possible.

PR #498, opened 2025-02-20
added get_delta_table_history

m-kovalsky:


src/sempy_labs/_delta_analyzer.py
```diff
@@ -154,29 +208,31 @@ def delta_analyzer(
...
+            # Loop through the columns
+            if column_stats:
+                for j in range(row_group.num_columns):
+                    column_chunk = row_group.column(j)
+                    total_compressed_size += column_chunk.total_compressed_size
+                    total_uncompressed_size += column_chunk.total_uncompressed_size
+
+                    # Generate Column Chunk Dataframe
+                    new_data = {
+                        "ParquetFile": file_name,
```
eisber, line 220: to be consistent with other data frames I'd suggest to use spaces. e.g. "Parquet File"

src/sempy_labs/_delta_analyzer.py
```diff
@@ -214,55 +274,58 @@ def delta_analyzer(
...
+        )
+        column_df = column_chunk_df.groupby(
+            ["ColumnName", "ColumnType"], as_index=False
+        ).agg({"CompressedSize": "sum", "UncompressedSize": "sum"})
+
+        # Add distinct count to column_df
+        for ind, r in column_df.iterrows():
+            col_name = r["ColumnName"]
+            if approx_distinct_count:
+                dc = _get_column_aggregate(
```
eisber, line 295: why call the function twice? just set the function parameter to either approx or distinctcount?

src/sempy_labs/_delta_analyzer.py
```diff
@@ -307,3 +370,52 @@ def delta_analyzer(
...
+    )
+    path = create_abfss_path(lakehouse_id, workspace_id, table_name)
+
+    from delta import DeltaTable
+
+    delta_table = DeltaTable.forPath(spark, path)
+    delta_table = DeltaTable.forPath(spark, path)
+    df = delta_table.history().toPandas()
+
+    df.rename(columns=lambda col: camel_to_title(col), inplace=True)
```
eisber, line 419: why rename the column names? aren't the column name ids, in the sense of a user going to use them else where (e.g. a query which won't work)?

eisber (changes requested):

eisber (approved):

PR #499, opened 2025-02-21
Handle optional columns in executeQueries API call

grenzi:
if the first returned row of a query doesn't contain all possible columns, the current code creates an incorrect dataframe. 

description of row keys() length for one of my queries:
count: 13586.
mean: 10.09
std: 0.97
min: 7.0
25%: 9.0
50%: 10.0
75%: 11.0
max: 11.0

this fix builds a complete column list before building the resultant dataframe, backfilling with none


grenzi:

m-kovalsky (approved):
Nice, thanks!

PR #501, opened 2025-02-22
Added ability to target all Fabric data store types with create_shortcut_onelake

crazy-treyn:
Initially the `create_shortcut_onelake` function only allowed targeting a Lakehouse as a source. However, the APIs being used to create shortcuts can target any `itemId`.

This PR includes updates to support creating shortcuts to any kind of Onelake data store item type; Lakehouse, Warehouse, MirroredDatabase, KQLDatabase, and SQLDatabase.

**Note**: The parameter structure of the function had to change slightly, including names of parameters and a new parameter for the data store type. This will break existing implementations using the original parameter structure of the `create_shortcut_onelake` function.

### DIFF
--- a/src/sempy_labs/lakehouse/_shortcuts.py
+++ b/src/sempy_labs/lakehouse/_shortcuts.py
@@ -1,5 +1,6 @@
 import sempy.fabric as fabric
 from sempy_labs._helper_functions import (
+    resolve_item_id,
     resolve_lakehouse_name,
     resolve_lakehouse_id,
     resolve_workspace_name_and_id,
@@ -13,7 +14,8 @@ from uuid import UUID
 
 def create_shortcut_onelake(
     table_name: str,
-    source_lakehouse: str,
+    source_item: str | UUID,
+    source_item_type: str,
     source_workspace: str | UUID,
     destination_lakehouse: str,
     destination_workspace: Optional[str | UUID] = None,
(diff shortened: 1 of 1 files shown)


crazy-treyn:

src/sempy_labs/lakehouse/_shortcuts.py
```diff
@@ -13,7 +14,8 @@
 
 def create_shortcut_onelake(
     table_name: str,
-    source_lakehouse: str,
+    source_item: str | UUID,
```
m-kovalsky, line 17: let's add **kwargs and then set if source_item is not entered, take 'source_lakehouse' from the kwargs. This will ensure that the function doesn't break for current users. Also, let's have the source_item_type default to 'Lakehouse'. This will also help with not breaking existing functionality.

m-kovalsky: See here for an example of kwargs in action to save the existing functionality from breaking

https://github.com/microsoft/semantic-link-labs/blob/b857010de39a6b0ca308e786ac011ca0285cd01f/src/sempy_labs/_capacities.py#L278

m-kovalsky (changes requested):

crazy-treyn (commented):
Good suggestion. I made the updates, let me know if you think it needs anything else. Cheers

src/sempy_labs/lakehouse/_shortcuts.py
```diff
@@ -58,13 +71,21 @@ def create_shortcut_onelake(
...
+        )
 
     (source_workspace_name, source_workspace_id) = resolve_workspace_name_and_id(
         source_workspace
     )
-    source_lakehouse_id = resolve_lakehouse_id(source_lakehouse, source_workspace_id)
-    source_lakehouse_name = fabric.resolve_item_name(
-        item_id=source_lakehouse_id, type="Lakehouse", workspace=source_workspace_id
+
+    source_item_id = resolve_item_id(source_item, source_workspace_id)
```
m-kovalsky, line 86: it's probably easier if you use 'resolve_item_name_and_id' from the helper functions.

m-kovalsky (changes requested):
one comment for making code cleaner

src/sempy_labs/lakehouse/_shortcuts.py
crazy-treyn: Updated. I also made a similar change resolving the name and id of the destination lakehouse using the same `resolve_item_name_and_id` function for consistency.

crazy-treyn:
Updated. I also made a similar change resolving the name and id of the destination lakehouse using the same `resolve_item_name_and_id` function for consistency.

crazy-treyn:
@m-kovalsky I think this is ready for merge, but please let me know if there's anything else I need to change.

m-kovalsky:
Done in [0.12.9](https://github.com/microsoft/semantic-link-labs/releases/tag/0.12.9).

PR #508, opened 2025-02-24
create item schedules

m-kovalsky:


src/sempy_labs/_job_scheduler.py
```diff
@@ -213,3 +213,227 @@ def run_on_demand_item_job(
...
+
+
+def create_item_schedule_cron(
+    item: str | UUID,
+    type: str,
+    start_date_time: str,
+    end_date_time: str,
+    local_time_zone: str,
+    job_type: str = "DefaultJob",
+    interval: int = 10,
```
eisber, line 225: suggestion: either use pd.TimeDelta  or call it internal_minutes?

m-kovalsky: updated to interval_minutes

PR #517, opened 2025-02-27
improve list_shortcuts

lBilali:
I took the liberty to change a bit list_shortcuts function
let me know is this PR does more than you would like.

The list of changes are:
1. Add optional path parameter to list_shortcuts
2. Add an extra column "Source Properties Raw" to the result
This can be handy if a mapping is missing 
for more it can be used when crating before creating a new shortcut if it already exists and has the same target. I think this will make it much easier to compare
3. Improve performance by caching artifacts of the target workspaces
I run a test with over 100 shortcuts coming from the same workspace and it took around 4 min, while this version run in 2 seconds
4. clean up a bit how the mapping is done. make it a bit more generic and easier to add new targets latter

P.S. I think this function should be moved under `/lakehouse/_shortcuts` where all other shortcut functions live. a bit more natural location in my opinion :)


m-kovalsky (approved):
Looks good, thanks! Somehow this function was missed at the initial reorganization of functions into subpackages. You are correct that this should be in the lakehouse subpackage. I'll move it over there but keep it in sempy_labs for some time to allow folks to make the change without breaking their code.

PR #544, opened 2025-03-06
offer pure python option for optimize

m-kovalsky:


a-holm:
Thanks for this significant contribution towards enabling pure Python execution for Delta operations! The use of helper functions in `_helper_functions.py` to abstract the execution environment is a great approach.

A few points (that I do for practice):

1.  **Documentation:** Several new helper functions (`_get_aggregate`, `_read_delta_table`, etc.) and updated functions/parameters (`save_as_delta_table`, `create_abfss_path`, `get_delta_table_history`) need docstrings or updates. It is so difficult for new people to read code without docstrings.
2.  **`delta_analyzer` Restriction:** The change in `_delta_analyzer.py` explicitly prevents it from running in a pure Python environment. Could you clarify the reasoning for this, as the PR's goal is generally to *add* pure Python support? Perhaps adding a note to its docstring explaining the Spark dependency?
3.  **`get_lakehouse_tables` Cross-Workspace:** The diff removes a check that previously restricted `count_rows=True` to the default lakehouse/workspace. Does the new implementation using helpers now support this cross-workspace, or was the check removed for other reasons? 
4.  **Vacuum Configuration:** The `spark.conf.set("spark.databricks.delta.vacuum.parallelDelete.enabled", "true")` line was removed in `vacuum_lakehouse_tables`. Does the `deltalake` vacuum implementation handle this implicitly, or could this affect behavior/performance when running vacuum in a Spark environment via this function?
5.  **Dependencies:** This change introduces `deltalake` and `polars` as new dependencies for the pure Python path. We'll need to ensure these are added to the project setup. I can't find any other mentions of this anywhere else in the repo, you probably know what you are doing. Just mentioning it just in case.

Overall, this seem to be a valuable direction for the library!

Issue #547, opened 2025-03-08
Delta Analyzer doesn't work well with schema-enabled lakehouse


dbrownems:
**Describe the bug**
The 

**To Reproduce**
In a schema enabled lakehouse neither 

table_name = 'FactInternetSales' # Enter the name of the delta table

nor

table_name = 'dbo.FactInternetSales' # Enter the name of the delta table

work.  You must use 

table_name = 'dbo/FactInternetSales' # Enter the name of the delta table

instead.

m-kovalsky:
Thanks David. Made this [PR](https://github.com/microsoft/semantic-link-labs/pull/548) for the next release.

m-kovalsky:
Available in [0.9.6](https://github.com/microsoft/semantic-link-labs/releases/tag/0.9.6).

PR #548, opened 2025-03-09
added schema parameter for issue 547

m-kovalsky:
#547


src/sempy_labs/_delta_analyzer.py
```diff
@@ -46,6 +46,7 @@ def get_parquet_file_infos(path):
 @log
 def delta_analyzer(
     table_name: str,
+    schema: Optional[str] = None,
```
eisber, line 49: can this be a fix in semantic-link?
also this breaks API compat, you'd have to version bump a major version as this break positional arguments.

eisber (approved):

Issue #448, opened 2025-02-06
Can semantic link labs be used to get a manually-created semantic model to update schema based on LH tables


gefawcet:
### Can semantic link labs be used to get a manually-created semantic model to update schema based on LH tables?

Trying to help a customer who wants to refresh a manually-created semantic model based on lakehouse tables **such that it adds or drops columns from the model based on changes to the lakehouse**.

This is the behavior we get in the default semantic model. (model schema auto-updates per LH schema)

In a manual semantic model, the only way we know to get the same to happen is through the service UI (going to "Edit model > Refresh").  That will pick up LH changes.  Refresh alone (called via API or from a Pipeline) won't do it.

They would like a way to call something in code that would execute the same so that model updates schema.

**Is there a way to use semantic link labs to do this?**

m-kovalsky:
See this function:

https://semantic-link-labs.readthedocs.io/en/stable/sempy_labs.directlake.html#sempy_labs.directlake.direct_lake_schema_sync

gefawcet:
I see the add_to_model  flag will add columns.

What about removing columns that have been deleted from the LH ?

Is there a way to set an option for the function to do that?

m-kovalsky:
Available in [0.9.6](https://github.com/microsoft/semantic-link-labs/releases/tag/0.9.6).

PR #549, opened 2025-03-10
added remove_from_model to direct_lake_schema_sync

m-kovalsky:
#448


src/sempy_labs/_helper_functions.py
```diff
@@ -1558,6 +1558,7 @@ def _convert_data_type(input_data_type: str) -> str:
         "date": "DateTime",
         "double": "Double",
         "float": "Double",
+        "binary": "Boolean",
```
eisber, line 1561: is binary really boolean or is it a binary stream (like bytes?)

eisber (approved):

PR #553, opened 2025-03-10
make destination lakehouse optional when creating onelake shortcuts

lBilali:
This PR makes destination lakehouse optional which will resolve to attached lakehouse on the notebook

as is today you can get the same by providing `destination_lakehouse=None` 
This PR makes it official and documented

Improve a bit the error handling if no lakehouse is attached or no ID or Name are provided


m-kovalsky:
Thanks. This will actually be resolved with this [PR](https://github.com/microsoft/semantic-link-labs/pull/550).

PR #560, opened 2025-03-12
Add format parameter when receiving/creating or updating notebook definitions

lBilali:
adding format parameter for get_notebook_definition, create_notebook and update_notebook_definition

when format parameter is not provided  the notebook definition should be GIT friendly 
if format = 'ipynb' is provided then the definition should be standard ipynb which is a json file

fixed  issues with encoding  the content within create_notebook and update_notebook_definition functions

> create_notebook  now by default has format=None which is a braking change


src/sempy_labs/_notebooks.py
```diff
@@ -20,13 +20,21 @@
 
 
 def _get_notebook_definition_base(
-    notebook_name: str, workspace: Optional[str | UUID] = None
+    notebook_name: str,
+    workspace: Optional[str | UUID] = None,
+    format: Optional[str] = None
```
m-kovalsky, line 25: What other formats besides '.ipynb' are supported here? I didn't add the 'format' parameter since it appears only '.ipynb' is supported - which is the default anyways. Please let me know if I'm mistaken.

m-kovalsky (changes requested):

src/sempy_labs/_notebooks.py
lBilali: Yes the APIs are not very well documented on this

but if no format is provided the APIs return or expect the notebook on .py GIT friendly format

lBilali: We can potentially change this to only work with standard .ipynb format or only with .py (GIT Friendly)
But as is today without this PR the functions are inconsistent
so you can't get the definition with the get_notebook_definition and then pass the result to create_notebook function

lBilali: here is getting the definition with and without format

lBilali:
Hei,
Any thoughts on this?

m-kovalsky:
> Hei, Any thoughts on this?

This will need to wait until after fabcon. Only bugfixes and features relevant to fabcon are being taken up now.

a-holm:
Thanks for adding the `format` parameter and fixing the encoding issues! This looks like a valuable addition for managing different notebook formats.

A few minor points:
*   You fixed a small, unrelated typo fix in `_helper_functions.py`. While harmless, mentioning minor cleanups like this in the description (or separating them) is good practice. Might make it easier for the reviewers.
* It looks like you also fixed a potential path issue in `create_notebook` and `update_notebook_definition` (changing `notebook-content..{type}` to `notebook-content.{type}`). Good catch! Just wanted to mention it, seems tricky to find.

lBilali:
hi @m-kovalsky ,
Did you have any time to look at this?

thanks

m-kovalsky (commented):
just a simple code-simplification. then will approve & merge.

m-kovalsky (approved):
thank you!

PR #563, opened 2025-03-16
M kovalsky/restorelaketable

m-kovalsky:


src/sempy_labs/_blobs.py
```diff
@@ -0,0 +1,96 @@
...
+        raise ValueError(
+            f"{icons.red_dot} Invalid container '{container}' within the file_path parameter. Expected 'Tables' or 'Files'."
+        )
+
+    response = _request_blob_api(
+        request=f"{workspace_id}/{lakehouse_id}/{container}?restype=container&comp=list&include=deleted"
+    )
+    root = ET.fromstring(response.content)
+    response_json = _xml_to_dict(root)
+    for blob in response_json["EnumerationResults"]["Blobs"]["Blob"]:
```
eisber, line 53: I'd suggest to be a bit more defensive here. response_json.get("EnumerationResults", {}).get(...)

src/sempy_labs/_blobs.py
```diff
@@ -0,0 +1,96 @@
...
+):
+
+    import requests
+    import notebookutils
+    from sempy.fabric.exceptions import FabricHTTPException
+
+    if isinstance(status_codes, int):
+        status_codes = [status_codes]
+
+    token = notebookutils.credentials.getToken("storage")
```
eisber, line 79: are you always using notebookutils to get the credentials or sometimes synapseml utils or sometimes SynapseTokenProvider() ?

eisber (changes requested):

src/sempy_labs/_blobs.py
m-kovalsky: always using notebookutils in this scenario

eisber (approved):

Issue #561, opened 2025-03-13
list_workspace_users() with Service Principle error


FrankPreusker:
SLL v0.9.6

Running list_workspace_users() with my own user ID is working ok

Running the same function with a Service Principle is producing an error

![Image](https://github.com/user-attachments/assets/462a847b-223e-4cb3-bbba-c96ee339bda6)

---------------------------------------------------------------------------
FabricHTTPException                       Traceback (most recent call last)
File /nfs4/pyenv-955aa317-b5c9-45b5-81f0-27dfd2e0f3a1/lib/python3.11/site-packages/sempy_labs/_helper_functions.py:847, in resolve_workspace_name(workspace_id)
    846 try:
--> 847     response = _base_api(
    848         request=f"/v1/workspaces/{workspace_id}", client="fabric_sp"
    849     ).json()
    850 except FabricHTTPException:

File /nfs4/pyenv-955aa317-b5c9-45b5-81f0-27dfd2e0f3a1/lib/python3.11/site-packages/sempy_labs/_helper_functions.py:1758, in _base_api(request, client, method, payload, status_codes, uses_pagination, lro_return_json, lro_return_status_code)
   1757 if method == "get":
-> 1758     response = c.get(request)
   1759 elif method == "delete":

File /nfs4/pyenv-955aa317-b5c9-45b5-81f0-27dfd2e0f3a1/lib/python3.11/site-packages/sempy/fabric/_client/_rest_client.py:155, in BaseRestClient.get(self, path_or_url, *args, **kwargs)
    137 """
    138 GET request to the Fabric and PowerBI REST API.
    139 
   (...)
    153     The response from the REST API.
    154 """
--> 155 return self.request("GET", path_or_url, *args, **kwargs)

File /nfs4/pyenv-955aa317-b5c9-45b5-81f0-27dfd2e0f3a1/lib/python3.11/site-packages/sempy/fabric/_client/_rest_client.py:323, in FabricRestClient.request(self, method, path_or_url, lro_wait, lro_max_attempts, lro_operation_name, *args, **kwargs)
    298 """
    299 Request to the Fabric REST API.
    300 
   (...)
    321     The response from the REST API.
    322 """
--> 323 response = super().request(method, path_or_url, *args, **kwargs)
    325 if not lro_wait or response.status_code != 202:

[...]

PR #565, opened 2025-03-17
fixed issue 561

m-kovalsky:
#561

### DIFF
--- a/src/sempy_labs/admin/_basic_functions.py
+++ b/src/sempy_labs/admin/_basic_functions.py
@@ -350,14 +350,40 @@ def list_workspace_access_details(
     return df
 
 
+def _resolve_workspace_name(workspace_id: Optional[UUID] = None) -> str:
+    from sempy_labs._helper_functions import _get_fabric_context_setting
+    from sempy.fabric.exceptions import FabricHTTPException
+
+    if workspace_id is None:
+        workspace_id = _get_fabric_context_setting(name="trident.workspace.id")
+
+    try:
+        workspace_name = (
+            _base_api(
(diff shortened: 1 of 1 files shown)


src/sempy_labs/admin/_basic_functions.py
```diff
@@ -350,14 +350,40 @@ def list_workspace_access_details(
     return df
 
 
+def _resolve_workspace_name(workspace_id: Optional[UUID] = None) -> str:
```
lBilali, line 353: sempy.fabric has already a function resolve_workspace_name with similar signature

Any good reason not to use that one but create your own?

src/sempy_labs/admin/_basic_functions.py
```diff
@@ -350,14 +350,40 @@ def list_workspace_access_details(
...
 def _resolve_workspace_name_and_id(
     workspace: str | UUID,
 ) -> Tuple[str, UUID]:
 
-    from sempy_labs._helper_functions import resolve_workspace_name_and_id
+    from sempy_labs._helper_functions import _get_fabric_context_setting
 
     if workspace is None:
-        (workspace_name, workspace_id) = resolve_workspace_name_and_id()
+        workspace_id = _get_fabric_context_setting(name="trident.workspace.id")
```
lBilali, line 382: here as well sempy.fabric has fabric.resolve_workspace_id

BTW this is me trying to understand it more then a code review :)

src/sempy_labs/admin/_basic_functions.py
m-kovalsky: I recreated these functions in labs so that they (and dependent functions) support Service Principal

eisber (approved):

PR #578, opened 2025-03-19
M kovalsky/dltoimportupgrade

m-kovalsky:


src/sempy_labs/directlake/_generate_shared_expression.py
```diff
@@ -78,4 +82,8 @@ def generate_shared_expression(
     end_expr = "\nin\n\tdatabase"
     mid_expr = f'Sql.Database("{sqlEPCS}", "{sqlepid}")'
 
-    return f"{start_expr}{mid_expr}{end_expr}"
+    # Build DL/OL expression
+    if not use_sql_endpoint and item_type == "Lakehouse":
+        return f'AzureDataLakeStorage{{"server":"onelake.dfs.fabric.microsoft.com","path":"/{workspace_id}/{item_id}/"}}'
```
eisber, line 87: I think the endpoint is different in daily/dxt/...

m-kovalsky: Zoe gave this to me

eisber: ok, can you check w/ Zoe on the endpoint for dxt/daily. I can already see that IDEAs will run into issues :)

m-kovalsky: Zoe tested and it also works in msit. I don’t think we need the prefix

bo-zhou-ms (approved):

PR #579, opened 2025-03-19
ecotte/Admin ExportDataflow and Capacity Refreshables

ecotte:
Implementing the following functions:

- Admin Export Dataflow 
- Admin Capacity Refreshables
- Update in the _helper_functions the function _update_dataframe_datatypes to accept "list" and "dict" as data type, and the list will evaluate if the list is empy, in that case return None.
- Update in _tenant the different "list settings" functions to return list in the columns that uses lists.


JeridInsight:
@ecotte  Your description says that you're wrapping the "Admin - Capacity Refreshables".  But this is the Get Refreshables endpoint, right?  I hope so.  Looks like you're waiting on a review.  Thanks for building this one,  I need it.  : )

a-holm:
Thanks for adding these admin functions! The implementation for `export_dataflow` and `get_refreshables` looks good and follows the established patterns.

A couple of points for discussion which might make it easier for the reviewers if you answer:

1.  **Pagination in `get_refreshables`:** The implementation in `src/sempy_labs/admin/_capacities.py` calls `_base_api` for the refreshables endpoint but doesn't seem to include the `uses_pagination=True` flag. Since the underlying API supports pagination, this might lead to incomplete results if there are many refreshables. Could we verify if pagination is needed and add the flag if necessary?
2.  **Scope:** The PR also includes helpful updates to `_helper_functions.py` (list/dict type handling) and `_tenant.py` (using the list type for security groups). While these are good improvements, they weren't mentioned in the PR description. It might be clearer to include these details or potentially move them to a separate PR focused on refactoring/bug fixes.
3.  **Testing:** Could you confirm if unit/integration tests have been added or updated for the new functions and the modified helpers/tenant functions? How do we do testing in this project?

Overall, great additions, just need to clarify the pagination and scope.

ecotte:
> 1. lead to incomplete results if there are many refreshables. Could we verify if pagination is needed and add the flag if necess

For the get_refreshables, it´s base on the API https://learn.microsoft.com/en-us/rest/api/power-bi/admin/get-refreshables-for-capacity, this one doesn´t support "continuationToken" and "continuationUri", so the "user_pagination" doesn´t apply. I know we can use pagination with "top" and "skip", and actually I have implemented that on the calling side in a case, but not as part of the SLL.

m-kovalsky (changes requested):

src/sempy_labs/admin/_dataflows.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@log
+def export_dataflow(
+    dataflow: str | UUID,
+) -> dict:
+    """
+    Shows a list of datasets for the organization.
+
+    This is a wrapper function for the following API: `Admin - Dataflows ExportDataflowAsAdmin <https://learn.microsoft.com/en-us/rest/api/power-bi/admin/dataflows-export-dataflow-as-admin>`_.
```
m-kovalsky, line 21: remove 'en-us'

src/sempy_labs/admin/_capacities.py
```diff
@@ -309,3 +310,135 @@ def list_capacity_users(capacity: str | UUID) -> pd.DataFrame:
     _update_dataframe_datatypes(dataframe=df, column_map=columns)
 
     return df
+
+
+@log
+def get_refreshables(
```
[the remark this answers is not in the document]
m-kovalsky: Hmm then maybe we will just stick with get. Honestly I am not so familiar with these APIs.

src/sempy_labs/admin/_capacities.py
```diff
@@ -309,3 +310,135 @@ def list_capacity_users(capacity: str | UUID) -> pd.DataFrame:
...
+    expand: Optional[str] = None,
+    filter: Optional[str] = None,
+    skip: Optional[int] = None,
+) -> pd.DataFrame:
+    """
+    Returns a list of refreshables for the organization within a capacity.
+
+    Power BI retains a seven-day refresh history for each dataset, up to a maximum of sixty refreshes.
+
+    This is a wrapper function for the following API: `Admin - Get Refreshables <https://learn.microsoft.com/rest/api/power-bi/admin/get-refreshables>`_.
```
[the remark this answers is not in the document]
ecotte: it now supports both APIs with an optional parameter 'capacity'

src/sempy_labs/admin/_capacities.py
```diff
@@ -309,3 +310,135 @@ def list_capacity_users(capacity: str | UUID) -> pd.DataFrame:
...
+                else None
+            ),
+            "Capacity Name": i.get("capacity", {}).get("displayName"),
+            "Capacity SKU": i.get("capacity", {}).get("sku"),
+            "Refresh Count": i.get("refreshCount", 0),
+            "Refresh Failures": i.get("refreshFailures", 0),
+            "Average Duration": i.get("averageDuration", 0),
+            "Median Duration": i.get("medianDuration", 0),
+            "Refreshes Per Day": i.get("refreshesPerDay", 0),
+            "Refresh Type": i.get("lastRefresh", {}).get("refreshType"),
```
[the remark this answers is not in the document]
ecotte: Done for 'lastRefresh' and 'refreshSchedule'

src/sempy_labs/admin/_dataflows.py
```diff
@@ -0,0 +1,42 @@
...
+    dataflow : str | UUID, default=None
+        The dataflow Name or Id.
+
+    Returns
+    -------
+    dict
+        Exported Json file.
+    """
+    if not _is_valid_uuid(dataflow):
+        dataflow = _resolve_item_id(item_name=dataflow, type="dataflow")
```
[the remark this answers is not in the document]
ecotte: Changed the _resolve_item_id to return the id directly if it´s a GUID so we dont need to query the items API eveytime the id needs to be resolved.

src/sempy_labs/admin/_capacities.py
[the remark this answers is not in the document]
ecotte: According to this discussion I left it as get_refreshables

src/sempy_labs/admin/_dataflows.py
ecotte: Fixed

src/sempy_labs/admin/_capacities.py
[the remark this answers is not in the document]
m-kovalsky: Let’s make a _resolve_capacity_id function (like the other resolve item id function) so we can potentially avoid running list_capacities where possible.

src/sempy_labs/admin/_items.py
```diff
@@ -21,16 +21,21 @@ def _resolve_item_id(
...
-        )
+        if len(dfI_filt) == 0:
+            raise ValueError(
+                f"The '{item_name}' {type} does not exist within the '{workspace}' workspace or is not of type '{type}'."
+            )
+
+        item_id = dfI_filt["Item Id"].iloc[0]
 
-    return dfI_filt["Item Id"].iloc[0]
+    return UUID(item_id, version=4)
```
m-kovalsky, line 38: Should return a string with thr UUID value, not a UUID data type

src/sempy_labs/admin/_items.py
```diff
@@ -21,16 +21,21 @@ def _resolve_item_id(
     type: Optional[str] = None,
     workspace: Optional[str | UUID] = None,
 ) -> UUID:
+    if _is_valid_uuid(item_name):
```
m-kovalsky, line 24: Since this is an internal function we should change the parameter name to ‘item’ instead of ‘item_name’. Item accepts either item id or item name.

src/sempy_labs/admin/_capacities.py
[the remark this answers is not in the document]
ecotte: Added _resolve_capacity_id

src/sempy_labs/admin/_items.py
ecotte: changed to return the basic str

src/sempy_labs/admin/_items.py
ecotte: did it, and validated the references.

src/sempy_labs/admin/_capacities.py
```diff
@@ -309,3 +328,147 @@ def list_capacity_users(capacity: str | UUID) -> pd.DataFrame:
...
+    capacity_id = None
+
+    if capacity is not None:
+        capacity_id = _resolve_capacity_id(capacity=capacity)
+
+    params = {}
+    url = (
+        "/v1.0/myorg/admin/capacities/refreshables"
+        if capacity_id is None
+        else f"/v1.0/myorg/admin/capacities/{capacity_id}/refreshables"
```
m-kovalsky, line 410: Can put the resolve capacity id here to make the code cleaner

ecotte: did the change, let me know what you think! :)

src/sempy_labs/admin/_dataflows.py
```diff
@@ -0,0 +1,40 @@
...
+from sempy_labs.admin._items import (
+    _resolve_item_id,
+)
+from uuid import UUID
+from sempy._utils._log import log
+
+
+@log
+def export_dataflow(
+    dataflow: str | UUID,
```
m-kovalsky, line 15: i think you will need a workspace parameter. otherwise _resolve_item_id will get confused as it takes a workspace parameter.

m-kovalsky (changes requested):

src/sempy_labs/admin/_dataflows.py
ecotte: Added as an optional parameter for export_dataflow.

Also added the logic in _resolve_item_id to look into attached lakehouse or notebook workspace in case that a DF name is provided and not the WS.

Corrected the text of list_items to specify that if WS is None, it looks into all the WS.

m-kovalsky (approved):
Thanks Edgar!

PR #580, opened 2025-03-20
Added return_df parameter to return the result as a df

AkhilGrandhi:
Added a return_df parameter of boolean type. If set to True, the function will return a DataFrame containing bpa_rules.

### DIFF
--- a/src/sempy_labs/_model_bpa_bulk.py
+++ b/src/sempy_labs/_model_bpa_bulk.py
@@ -28,6 +28,7 @@ def run_model_bpa_bulk(
     workspace: Optional[str | UUID | List[str | UUID]] = None,
     skip_models: Optional[str | List[str]] = ["ModelBPA", "Fabric Capacity Metrics"],
     skip_models_in_workspace: Optional[dict] = None,
+    return_df: bool = False,
 ):
     """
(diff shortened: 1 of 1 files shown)


AkhilGrandhi:

a-holm:
Thanks for adding the `return_df` feature! Being able to get the results back as a DataFrame programmatically is a useful addition.

However, I've identified a couple of issues with the current implementation:

1.  **Premature Return:** The `return df` statement is currently placed inside the main loop over workspaces (`src/sempy_labs/_model_bpa_bulk.py`, around line 190 in the diff). This means if `return_df=True`, the function will exit after processing and saving results for only the *first* workspace, instead of collecting results from all specified workspaces (if I understand the code correctly). To return the complete results, the aggregation logic needs adjustment (e.g., accumulate all results into a master DataFrame) and the `return` statement should be moved *after* the workspace loop finishes.
2.  **Missing Documentation:** The new `return_df` parameter needs to be added to the function's docstring to explain its purpose and usage.

Could you please adjust the logic to ensure all workspace results are collected before returning when `return_df=True` and update the docstring accordingly? It would probably make it easier for the microsoft fellas.

AkhilGrandhi:
Hi @a-holm 

Thanks for the review! I have now updated the code with the correct logic and proper comments. Please review and apply the changes.

PR #583, opened 2025-03-20
ecotte/Fix on list_capacity_tenant_settings_overrides for Capacity Id

ecotte:
Fix when pulling all the capacities at the same time left the ID as blank.


m-kovalsky (approved):
Thanks!

PR #588, opened 2025-03-25
added list_blobs

m-kovalsky:


src/sempy_labs/lakehouse/_blobs.py
```diff
@@ -1,13 +1,160 @@
...
+        path_prefix = f"{workspace_id}/{lakehouse_id}"
+    else:
+        if container not in ["Tables", "Files"]:
+            raise ValueError(
+                f"{icons.red_dot} Invalid container '{container}' within the file_path parameter. Expected 'Tables' or 'Files'."
+            )
+        path_prefix = f"{workspace_id}/{lakehouse_id}/{container}"
+
+    response = _request_blob_api(
+        request=f"{path_prefix}?restype=container&comp=list&include=deleted"
```
eisber, line 94: is there any paging you need to take care of?

m-kovalsky: there is a 'NextMarker' but I'm not exactly sure how to use it. It's not really documented.

https://learn.microsoft.com/en-us/rest/api/storageservices/list-blobs?tabs=microsoft-entra-id

bo-zhou-ms (approved):

PR #604, opened 2025-04-03
create vpax

m-kovalsky:


src/sempy_labs/_vpax.py
```diff
@@ -0,0 +1,314 @@
...
+    assembly_path = current_dir / "dotnet_lib"
+    # rt = get_coreclr(
+    #    runtime_config=os.fspath(f"{assembly_path}/dotnet.runtime.config.json")
+    # )
+    # set_runtime(rt)
+
+    import clr
+
+    sys.path.append(os.fspath(assembly_path))
+    clr.AddReference(os.fspath(assembly_path / "Dax.Metadata.dll"))
```
eisber, line 45: can you pull these from nuget.org? I'd rather not have the binaries pushed to the git repo

src/sempy_labs/_vpax.py
```diff
@@ -0,0 +1,314 @@
...
+            f"{icons.warning} The .vpax file already exists at {path}. Set overwrite=True to overwrite the file."
+        )
+        return
+
+    vpax_stream = MemoryStream()
+    extractor_app_name = "VPAX Notebook"
+    extractor_app_version = "1.0"
+    column_batch_size = 50
+    token = notebookutils.credentials.getToken("pbi")
+    connection_string = f"data source=powerbi://api.powerbi.com/v1.0/myorg/{workspace_name};initial catalog={dataset_name};User ID=;Password={token};Persist Security Info=True;Impersonation Level=Impersonate "
```
eisber, line 144: is vpax nuget using adomd connection underneath? if so they should expose the token call back so that token refresh works.

eisber (changes requested):

eisber (approved):

PR #613, opened 2025-04-13
Update _generate_report.py

edwardpcharles:
adding the ability of the function to output to a notebooks built in resources - Would love to see this feature!

### DIFF
--- a/src/sempy_labs/report/_generate_report.py
+++ b/src/sempy_labs/report/_generate_report.py
@@ -2,6 +2,7 @@ import sempy.fabric as fabric
 import pandas as pd
 import json
 import os
+import base64
 from typing import Optional
 from sempy_labs._helper_functions import (
     resolve_workspace_name_and_id,
@@ -182,6 +183,7 @@ def get_report_definition(
     report: str | UUID,
     workspace: Optional[str | UUID] = None,
     return_dataframe: bool = True,
+    files_to_built_in_resources = False
 ) -> pd.DataFrame | dict:
     """
     Gets the collection of definition files of a report.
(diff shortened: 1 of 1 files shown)


edwardpcharles:

edwardpcharles:
@m-kovalsky I don't always want to have to attach a lakehouse to my notebook - I would much prefer to be able to save everything self-contained in the notebook.   I run a lot of little semantic link scripts across my models and keep them in notebooks in a specific workspace. 

In the case of this specific function - I want my team to be able to run a notebook and quickly download a report's definition, so they can grab 1 or 2 snippets of it and then basically dump the file in the recycling bin. However, if all the files end up on OneLake now every single one of those team members needs to also know how to interact with oneLake and not just run the script and work with the report definition file.

m-kovalsky:
How about we just make a helper function that saves any given item to the notebook resources? Seems more scalable than adding this parameter into each function.

m-kovalsky:
@edwardpcharles can have a versatile function like this:

```python
def save_to_file(file_path: str, content, file_type: Optional[str] = None):

    """
    Saves the content to a file in the specified format.

    Parameters
    ----------
    file_path : str
        The file path where the content will be saved.
        Example: "./build/myfolder/myfile.json" (will save to the Notebook resources)
    content : any
        The content to be saved.
    file_type : str, default=None
        The file type to save the content as. If None, it will be inferred from the file path.
    """

    import json
    from pathlib import Path
    import csv
    import sempy_labs._icons as icons

    file_path = Path(file_path)
    file_path.parent.mkdir(parents=True, exist_ok=True)

    if file_type is None:
        file_type = file_path.suffix.lower()

    if file_type == ".json":
        with open(file_path, "w") as f:
            json.dump(content, f, indent=4)
    elif file_type == ".txt":
        with open(file_path, "w") as f:
            f.write(str(content))
    elif file_type in [".csv", ".tsv"]:
        delimiter = "\t" if file_type == ".tsv" else ","
        with open(file_path, "w", newline="") as f:
            writer = csv.writer(f, delimiter=delimiter)
            writer.writerows(content)
    else:
        with open(file_path, "wb") as f:
            f.write(content)

    print(f"{icons.green_dot} The file has been saved to the following location: '{file_path}.")
```

edwardpcharles:
@m-kovalsky this would be great -- sorry for my delay in response work + life got in the way a bit!

PR #636, opened 2025-04-25
Updated the list_dataflows function to accept a list of workspaces

AkhilGrandhi:
Updated the list_dataflows function to accept a list of workspaces as a parameter. If the parameter is set to None, the function will now run for all workspaces by default. Removed Defaults to None which resolves to the workspace of the attached lakehouse
        or if no lakehouse attached, resolves to the workspace of the notebook.


m-kovalsky:
Will not be doing this. If you want to get data from multiple workspaces, simply run the function over the multiple workspaces. Changing this functionality would mean changing it in practically all functions which is not a priority.

PR #697, opened 2025-06-03
Update _dataflows.py

itsnotaboutthecell:
Updated several functions with the new Fabric APIs, updated several functions to delineate between Power BI and Fabric dataflow endpoint routing.


m-kovalsky:
closing

PR #706, opened 2025-06-09
Adding support for schema enabled Lakehouse in function get_lakehouse…

tarente:
Adding support for schema enabled Lakehouse in function get_lakehouse_tables until the Lakehouse schema is in GA.

### DIFF
--- a/src/sempy_labs/lakehouse/_get_lakehouse_tables.py
+++ b/src/sempy_labs/lakehouse/_get_lakehouse_tables.py
@@ -108,7 +108,7 @@ def get_lakehouse_tables(
                 new_data = {
                     "Workspace Name": workspace_name,
                     "Lakehouse Name": lakehouse_name,
-                    "Schema Name": "dbo",
+                    "Schema Name": "",
                     "Table Name": i.get("name"),
                     "Format": i.get("format"),
                     "Type": i.get("type"),
@@ -155,27 +155,18 @@ def get_lakehouse_tables(
             schema_name = r["Schema Name"]
             table_name = r["Table Name"]
             if r["Type"] == "Managed" and r["Format"] == "delta":
-                delta_table_path_with_schema = create_abfss_path(
-                    lakehouse_id, workspace_id, table_name, schema_name
+                delta_table_path = (
+                    create_abfss_path(
+                        lakehouse_id, workspace_id, table_name, schema_name
+                    )
+                    .replace("//", "/")             # When schema_name = ""
+                    .replace("abfss:/", "abfss://") # Put back the // after abfss:
                 )
-
-                delta_table_path_without_schema = create_abfss_path(
-                    lakehouse_id, workspace_id, table_name
-                )
-
-                delta_table_path = None
+                    
                 if _pure_python_notebook():
                     from deltalake import DeltaTable
-
-                    try:
-                        delta_table_path = delta_table_path_with_schema
-                        delta_table = DeltaTable(delta_table_path)
                     
-                    except Exception as e:
-                        use_schema = False
-                        delta_table_path = delta_table_path_without_schema
-                        delta_table = DeltaTable(delta_table_path)
-
+                    delta_table = DeltaTable(delta_table_path)
                     latest_files = [
                         file["path"]
                         for file in delta_table.get_add_actions().to_pylist()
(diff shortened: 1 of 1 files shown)


src/sempy_labs/lakehouse/_get_lakehouse_tables.py
```diff
@@ -83,35 +85,70 @@ def get_lakehouse_tables(
     if count_rows:  # Setting countrows defaults to extended=True
         extended = True
 
-    responses = _base_api(
-        request=f"v1/workspaces/{workspace_id}/lakehouses/{lakehouse_id}/tables",
-        uses_pagination=True,
-        client="fabric_sp",
-    )
+    ### Begin Schema Lakehouse
+    local_path = _mount(lakehouse=lakehouse_id, workspace=workspace_id)
```
m-kovalsky: this only needs to be run in case the API call fails

src/sempy_labs/lakehouse/_get_lakehouse_tables.py
```diff
@@ -83,35 +85,70 @@ def get_lakehouse_tables(
...
+
+        for r in responses:
+            for i in r.get("data", []):
+                new_data = {
+                    "Workspace Name": workspace_name,
+                    "Lakehouse Name": lakehouse_name,
+                    "Table Name": i.get("name"),
+                    "Format": i.get("format"),
+                    "Type": i.get("type"),
+                    "Location": i.get("location"),
```
m-kovalsky, line 115: let's add a 'Schema Name' column between Lakehouse Name & Table Name. For non-schema enabled lakehouses, the schema name should default to 'dbo'.

src/sempy_labs/lakehouse/_get_lakehouse_tables.py
```diff
@@ -83,35 +85,70 @@ def get_lakehouse_tables(
...
+                }
+                dfs.append(pd.DataFrame(new_data, index=[0]))
+    else:
+        # Get schema list
+        tablesPath = os.path.join(local_path, "Tables")
```
m-kovalsky: nit - please conform to the standard of using parameters which are all lowercase and with underscores. for example:
api_call; tables_path

src/sempy_labs/lakehouse/_get_lakehouse_tables.py
```diff
@@ -83,35 +85,70 @@ def get_lakehouse_tables(
...
+                tableName = table
+                schemaTableName = f"{schemaName}/{tableName}"
+                locationPath = f"abfss://{workspace_id}@onelake.dfs.fabric.microsoft.com/{lakehouse_id}/Tables/{schemaTableName}"
+                new_data = {
+                    "Workspace Name": workspace_name,
+                    "Lakehouse Name": lakehouse_name,
+                    "Table Name": schemaTableName,
+                    "Format": "delta",
+                    "Type": "Managed",
+                    "Location": locationPath
```
m-kovalsky: add the schema name column here too

src/sempy_labs/lakehouse/_get_lakehouse_tables.py
```diff
@@ -83,35 +85,70 @@ def get_lakehouse_tables(
...
+        # Iterate lstSchema
+        for schema in lstSchema:
+            schemaName = schema
+            # Get table list for schema
+            schemaTablePath = os.path.join(local_path, "Tables", schemaName)
+            lstTables = os.listdir(schemaTablePath)
+            for table in lstTables:
+                tableName = table
+                schemaTableName = f"{schemaName}/{tableName}"
+                locationPath = f"abfss://{workspace_id}@onelake.dfs.fabric.microsoft.com/{lakehouse_id}/Tables/{schemaTableName}"
```
m-kovalsky: use the create_abfss_path to simplify things and ensure the proper prefix to @onelake.dfs....

src/sempy_labs/lakehouse/_get_lakehouse_tables.py
```diff
@@ -83,35 +85,70 @@ def get_lakehouse_tables(
...
 
     if dfs:
         df = pd.concat(dfs, ignore_index=True)
 
     if extended:
         sku_value = get_sku_size(workspace_id)
         guardrail = get_directlake_guardrails_for_sku(sku_value)
-        local_path = _mount(lakehouse=lakehouse_id, workspace=workspace_id)
+        ### Begin Schema Lakehouse
+        # local_path = _mount(lakehouse=lakehouse_id, workspace=workspace_id)
```
m-kovalsky: make sure mount is called here (in case it is not called if the API is used).

m-kovalsky (changes requested):

src/sempy_labs/lakehouse/_get_lakehouse_tables.py
tarente: Be aware that this change implied further changes since the non-schema enabled Lakehouse will not show the files in (relative) path 'Tables/dbo/<table name>', but they are stored as 'Tables/<table name>'.

PR #717, opened 2025-06-17
Handling empty SQL endpoint refresh

vytautas9:
Currently, refreshing a SQL endpoint using Semantic Link Labs using `refresh_sql_endpoint_metadata` function fails when the target Lakehouse has no tables and we don't specify tables parameter in the function, for example:
```python
refresh_sql_endpoint_metadata(item = 'landing', type = 'Lakehouse', workspace = 'Fabric Workspace', tables = None)
```

This execution will fail with the following error message:
image
Error is caused by the fact that the API endpoint to refresh the SQL endpoint returns an empty response (but is a successful) when the Lakehouse is empty so the dataframe `df` will be empty and will cause the index issue when trying to reorder the columns (which do not exist).

The expected behavior of refreshing the empty SQL endpoint would be a successful execution of the function.

The proposed changes check if the response is empty or not:
- if not empty: continue with old code
- if empty: create an empty dataframe for return value and print a message

An example of the new output:
semantic-link-labs-error-2

Semantic link labs version: `0.10.1`

### DIFF
--- a/src/sempy_labs/_sql_endpoints.py
+++ b/src/sempy_labs/_sql_endpoints.py
@@ -157,45 +157,52 @@ def refresh_sql_endpoint_metadata(
         "Error Message": "string",
     }
 
-    df = pd.json_normalize(result)
+    if result:
+        df = pd.json_normalize(result)
 
-    # Extract error code and message, set to None if no error
-    df["Error Code"] = df.get("error.errorCode", None)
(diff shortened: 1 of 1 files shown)


vytautas9:

vytautas9:
solves #719

m-kovalsky (approved):
Great - thanks!

PR #775, opened 2025-07-23
listitemlabels

m-kovalsky:


src/sempy_labs/_labels.py
```diff
@@ -0,0 +1,124 @@
...
+
+    client = fabric.PowerBIRestClient()
+    response = client.get("/v1.0/myorg/capacities")
+    if response.status_code != 200:
+        raise FabricHTTPException("Failed to retrieve URL prefix.")
+    context = response.json().get("@odata.context")
+    prefix = context.split("/v1.0")[0]
+
+    response = requests.post(
+        f"{prefix}/metadata/informationProtection/artifacts",
```
eisber, line 98: is this a public endpoint? what scopes does it need?

PR #810, opened 2025-08-13
sp update funcs

m-kovalsky:


src/sempy_labs/_workspaces.py
```diff
@@ -320,7 +327,7 @@ def list_workspace_role_assignments(
         A pandas dataframe showing the members of a given workspace and their roles.
     """
 
-    (workspace_name, workspace_id) = resolve_workspace_name_and_id(workspace)
+    workspace_id = resolve_workspace_id(workspace)
```
ekote, line 330: if the workspace argument is ever None (default), what guarantees the workspace context is unambiguously set? Some downstream API endpoints may fail non-obviously (e.g., 404 or 401) if the implicit workspace resolution doesn’t match the user expectation, especially when running in async or multi-tenant contexts. Add assertion/logical guards for workspace_id presence before hitting the backend.

m-kovalsky: If the workspace parameter is None, it resolves to the workspace of the default lakehouse attached to the notebook. If no lakehouses are attached to the notebook then it resolves to the workspace in which the notebook resides.

PR #817, opened 2025-08-21
add Lakehouse Table Maintenance Example

MikeCarlo:
add sample notebook for Lakehouse maintenance

### DIFF
--- a/src/sempy_labs/_notebooks.py
+++ b/src/sempy_labs/_notebooks.py
@@ -163,7 +163,7 @@ def import_notebook_from_web(
     if len(dfI_filt) == 0:
         create_notebook(
             name=notebook_name,
-            notebook_content=response.content,
+            notebook_content=response.content.decode('utf-8'),
             workspace=workspace_id,
             description=description,
             format="ipynb",
@@ -216,7 +216,7 @@ def create_notebook(
         Defaults to None which places the notebook in the root of the workspace.
     """
 
-    notebook_payload = base64.b64encode(notebook_content).decode("utf-8")
+    notebook_payload = base64.b64encode(notebook_content.encode('utf-8')).decode("utf-8")
 
     definition_payload = {
         "parts": [


notebooks/Lakehouse Table Maintenance.ipynb
```diff
@@ -0,0 +1,405 @@
+{
+ "cells": [
+  {
```
m-kovalsky: 'Docs for Table Maintenance' should have a link?

notebooks/Lakehouse Table Maintenance.ipynb
```diff
@@ -0,0 +1,405 @@
...
+   "cell_type": "markdown",
+   "id": "5c27dfd1-4fe0-4a97-92e6-ddf78889aa93",
```
m-kovalsky: the vacuum and optimize functions work in both pyspark and pure python notebooks (pure python leverages the delta-rs library (deltalake). the guidance in this notebook just mentions pyspark.

notebooks/Lakehouse Table Maintenance.ipynb
```diff
@@ -0,0 +1,405 @@
...
+   "metadata": {
+    "nteract": {
```
m-kovalsky: nit - all parameter names should use underscores instead of snake case. as an example: lakehouse_name instead of lakehouseName. Or, you can just have a parameter of 'lakehouse' and mention that it can be either the name or ID.

notebooks/Lakehouse Table Maintenance.ipynb
```diff
@@ -0,0 +1,405 @@
...
+     "transient": {
+      "deleting": false
+     }
```
m-kovalsky: i find the row index and listing a bit odd and confusing. i don't think it's a realistic scenario. folks should simply enter in a list of tables to optimize/vacuum.

should also add an example of vacuuming using the retention days parameter.

Issue #841, opened 2025-09-03
Function list_capacities() is returning only single capacity


SrujanGirish:
**Describe the bug**
As a capacity admin for all 18 capacities, as well as a tenant admin, I've noticed that when I execute the `list_capacities()` function under admin, it only returns one capacity at a time. Each time I run the function, it returns a different capacity, despite the fact that we have a total of 18 capacities.

**To Reproduce**
1. In new notebook, install semantic link labs
2. Import `sempy_labs` module
3. Run the following command: `sempy_labs.admin.list_capacities()`

**Expected behavior**
Function should return the dataframe containing all the Capacities. 

**Screenshots**
Screenshot 1:

<img width="982" height="418" alt="Image" src="https://github.com/user-attachments/assets/62e4d1c3-623e-4159-b96f-bf64ac1f9e2d" />

Screenshot 2:

<img width="1060" height="445" alt="Image" src="https://github.com/user-attachments/assets/8fa3b83c-bb1a-4ea0-9ac9-0bf61dd1a736" />

Screenshot 3:

![Image](https://github.com/user-attachments/assets/c7c3edcf-e6a5-4507-8c63-2be92d30c754)

m-kovalsky:
Thanks. A fix was made for this for the next release.

m-kovalsky:
This is fixed as of [0.12.1](https://github.com/microsoft/semantic-link-labs/releases/tag/0.12.1).

PR #842, opened 2025-09-03
Bugfix: Include all capacities in list_capacities output

MitchSS:
**Summary**
Fixed an issue where the admin function list_capacities returned only one row even when multiple capacities were available.

**Details**

- Root Cause:rows.append(row) was placed outside the inner loop, so only the last capacity from each response page was added to the DataFrame.
- Fix: Moved rows.append(row) inside the inner loop to ensure all capacities are included in the result.


m-kovalsky:
#841

PR #877, opened 2025-09-18
fix create_notebook

lBilali:
provided notebook_content on create_notebook should be a string

on import_notebook_from_web  the content was provided as bytearray instead

this PR fixes it and makes sure the create_notebook treats notebook_content as string


m-kovalsky (approved):
Thanks!

Issue #889, opened 2025-09-26
Bug / enhancment - graph.list_groups() do not support pagination


vytautas9:
**Describe the bug**
Using a current version of `graph` module and particularly `graph.list_groups()` function is not feasible as the function does not support pagination and returns only top 100 groups.

**To Reproduce**
Execute:
``` python
with labs.service_principal_authentication(
    key_vault_uri=keyvault_url, 
    key_vault_tenant_id=tenant_id_secret_name,
    key_vault_client_id=client_id_secret_name,
    key_vault_client_secret=client_secret_secret_name):

    groups = graph.list_groups()
```

and check the number of groups:
``` python
len(groups)
```

the output will be 100 (if a tenant has >100 groups).

**Expected behavior**
`graph` module functions performs pagination on the back-end OR gives a user opportunity to perform cursor pagination (where applicable). This should apply to other `graph` module functions as well.

- Semantic link labs version: 0.12.3

**Additional context**
Will provide a PR with automatic pagination implementation as a proposed fix.

KimTutein:
We have observed the same problem. 

Moreover - the problem is the same with the list_users. The example below allso only load the first 100 users.

```
with labs.service_principal_authentication(
    key_vault_uri=key_vault_uri, 
    key_vault_tenant_id=key_vault_tenant_id,
    key_vault_client_id=key_vault_client_id,
    key_vault_client_secret=key_vault_client_secret):

    df = graph.list_users()
print(f"Number of users {len(df)}")
```

m-kovalsky:
Thanks, will get to this when I’m back from vacation.

m-kovalsky:
Fixed in [0.12.4](https://github.com/microsoft/semantic-link-labs/releases/tag/0.12.4).

PR #890, opened 2025-09-26
Graph API pagination

vytautas9:
Using a current version of `graph` module and particularly `graph.list_groups()` function is not feasible as the function does not support pagination and returns only top 100 groups.

This PR:
1. Adds a new `graph_pagination` function to the `_helper_functions.py` file which handles the graph api specific pagination (using odata.nextLink)
2. In the `_base_api` function, adds conditional statement to check if the client is `graph`, then perform graph api specific pagination, otherwise - standard Fabric one
3. Adjusts list functions in `graph` module to include pagination handling

P.S. *Could not test `list_sensitivity_labels` function so unsure if that one also requires the pagination handling.*

### DIFF
--- a/src/sempy_labs/_helper_functions.py
+++ b/src/sempy_labs/_helper_functions.py
@@ -1619,6 +1619,27 @@ def pagination(client, response):
     return responses
 
 
+def graph_pagination(response, headers):
+
+    responses = []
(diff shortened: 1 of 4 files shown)


vytautas9:
Solves #889

Issue #888, opened 2025-09-25
Option for lakehouse optimize/vacuum functions to exclude shortcuts


philosophicles:
**Is your feature request related to a problem? Please describe.**

Currently when running the functions lakehouse.optimize_lakehouse_tables() and lakehouse.vacuum_lakehouse_tables() in a pure-python notebook, I've been running into weird low-level errors from the deltalake package. My goal was to write a notebook that optimises all lakehouses in all workspaces in either our DEV environment, or our PRD environment; there are naturally some lakehouses here that have shortcuts to tables in other lakehouses, in other workspaces. The workspaces are approximately a medallion architecture, i.e. there's a directional flow of data from one workspace to another.

(We also have lakehouses with shortcuts to tables in _warehouses_, hence #886, but I'm leaving that part out of this suggestion.) 

The low-level errors seem intermittent as far as my testing goes, and _seem_ (but I'm not 100% sure of this) to relate only to lakehouses with shortcuts to other lakehouses. I'm not at all sure that these errors are really the fault of sempy-labs; they _seem_ to be avoidable if I optimize the lakehouses in the right order, so the bronze ones first, then silver ones that shortcut to bronze, and so on. Although, that could have just been good luck, since the whole thing is intermittent anyway. 

Just in case this is helpful, this is the start of such a low-level error: 

```
DeltaError: Failed to parse parquet: Parquet error: AsyncChunkReader::get_bytes error: Generic MicrosoftAzure error: Client error with status 403 Forbidden: <?xml version="1.0" encoding="utf-8"?><Error><Code>AuthenticationFailed</Code><Message>Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
RequestId:**<redacted GUID>**
Time:2025-09-25T16:55:26.7397587Z</Message><AuthenticationErrorDetail>Signature did not match. String to sign used was racwl
```

after which there is a spew of further unlabelled output, guids, timestamps, paths to _delta_log files, etc. The path does tell me which shortcut/table the error is happening on, but it's different each time, and if I run the optimize function again for _just_ that table, it usually works without error. It doesn't seem like it's one corrupted Delta table or something like that. 

**Describe the solution you'd like**

I think this can be side-stepped if the optimize and vacuum functions either: 

1. provide an extra boolean parameter allowing the caller to exclude shortcuts from being processed, if the caller wishes; or
[...]

PR #894, opened 2025-10-05
888

m-kovalsky:
#888


philosophicles:
Just to repeat my comment from #888, I think the line that creates df_filtered should be:

```
df_filtered = df_delta[~df_delta["FullPath"].isin(shortcuts)].reset_index(drop=True)
```

i.e. `df_filtered` should be built on `df_delta`, not on `df`. The FullPath column isn't present in `df`.

PR #916, opened 2025-10-20
M kovalsky/discover dataflow parameters

m-kovalsky:


src/sempy_labs/_dataflows.py
```diff
@@ -10,18 +10,25 @@
...
 import json
+from sempy._utils._log import log
 
 
+@log
 def list_dataflows(workspace: Optional[str | UUID] = None):
     """
     Shows a list of all dataflows which exist within a workspace.
 
+    This is a wrapper function for the following API: `Items - Create Dataflow <https://learn.microsoft.com/rest/api/fabric/dataflow/items/create-dataflow>`_.
```
bo-zhou-ms, line 28: Should be [Items - List Dataflows](https://learn.microsoft.com/en-us/rest/api/fabric/dataflow/items/list-dataflows?tabs=HTTP) here?

PR #917, opened 2025-10-20
Support for Lakehouses with Shared Column Names + Column Attribute Fixes for Fabric Compatibility

Alex-Real-USF:
This PR introduces two key enhancements:

Support for Lakehouses with Overlapping Column Names and Enables semantic model generation for Lakehouses that contain tables with identical column names.

Fix for Columns Misidentified as Measures in Microsoft Fabric.
Adds explicit column attributes to prevent Fabric from incorrectly interpreting certain columns as measures.

### DIFF
--- a/src/sempy_labs/directlake/_dl_helper.py
+++ b/src/sempy_labs/directlake/_dl_helper.py
@@ -180,16 +180,27 @@ def generate_direct_lake_semantic_model(
         for t in lakehouse_tables:
             tom.add_table(name=t)
             tom.add_entity_partition(table_name=t, entity_name=t, schema_name=schema)
-            dfLC_filt = dfLC[dfLC["Table Name"] == t]
+            dfLC_filt = dfLC.loc[dfLC["Table Name"] == t].drop_duplicates("Column Name")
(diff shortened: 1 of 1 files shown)


src/sempy_labs/directlake/_dl_helper.py
```diff
@@ -180,16 +180,27 @@ def dyn_connect():
         for t in lakehouse_tables:
             tom.add_table(name=t)
             tom.add_entity_partition(table_name=t, entity_name=t, schema_name=schema)
-            dfLC_filt = dfLC[dfLC["Table Name"] == t]
+            dfLC_filt = dfLC.loc[dfLC["Table Name"] == t].drop_duplicates("Column Name")
             for i, r in dfLC_filt.iterrows():
                 lakeCName = r["Column Name"]
                 dType = r["Data Type"]
                 dt = _convert_data_type(dType)
+                summarize_by = "Sum" if dt in ["Int64", "Double", "Decimal"] else "None"
```
m-kovalsky, line 188: else for all these should be None, not "None"

Issue #925, opened 2025-10-27
Add Department and Manager to graph.list_users() and graph.get_user()


KimTutein:
**Is your feature request related to a problem? Please describe.**
We love we can get a list of our users with graph.list_users. However the way we might use this list is to combine with for instance logs on our core semantic models. There we may se who initiated a call. We then want an overview / high level where we may add up on different departments. Having the department name and manager name we migt make a hirarchi from that which will help.

**Describe the solution you'd like**
For the graph.list_users() and graph.get_user() add the field Department and Manager in output

**Describe alternatives you've considered**
We could make the api call ourselfes but semantic link labs just makes it easier for us if added there.

**Additional context**
(none)

m-kovalsky:
Hi Kim - I'd be happy to add those in but I don't see them in the [API documentation](https://learn.microsoft.com/en-us/graph/api/user-list)? Could you confirm the output of the API and how it appears (as in the example from the API documentation below)?

{
    "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#users",
    "value": [
        {
            "businessPhones": [],
            "displayName": "Conf Room Adams",
            "givenName": null,
            "jobTitle": null,
            "mail": "Adams@contoso.com",
            "mobilePhone": null,
            "officeLocation": null,
            "preferredLanguage": null,
            "surname": null,
            "userPrincipalName": "Adams@contoso.com",
            "id": "6ea91a8d-e32e-41a1-b7bd-d2d185eed0e0"
        },
        {
            "businessPhones": [
                "425-555-0100"
            ],
            "displayName": "MOD Administrator",
            "givenName": "MOD",
            "jobTitle": null,
            "mail": null,
            "mobilePhone": "425-555-0101",
            "officeLocation": null,
            "preferredLanguage": "en-US",
            "surname": "Administrator",
            "userPrincipalName": "admin@contoso.com",
            "id": "4562bcc8-c436-4f95-b7c0-4f8ce89dca5e"
        }
    ]
}

KimTutein:
Hi @m-kovalsky 

I think this will do the trick:
https://graph.microsoft.com/v1.0/users?$select=id,mailNickname,jobTitle,mail,displayName,department&$expand=manager($select=displayName,id,mailNickname,mail,jobTitle)


You should get respons like this (changed with XXX some of the places) :
{
[...]

PR #926, opened 2025-10-28
925

m-kovalsky:
#925


KimTutein:
Hi @m-kovalsky 
I would love to test but to be frank I am not sure how I may install this version of semantic link labs in a notebook. Is there a sort of pip install I can make in a notebook or do I need to package this code as a pyhon file. Sorry - but the only things I have done so far is make small notebook with classes and saving as python files (then adding to a environment). 

If you can give me some guidance I will be happy to test.

Issue #940, opened 2025-11-01
list_item_access_details ignores item types and fails to list semantic models


muhssamy:
**Describe the bug**
running on python notebook on fabric after listing the items using list_items then when iterating to find the items permissions it fails on semantic model showing it is not found. i tested it from postman also it has the same issue
after adding ?type=SemanticModel in postman it works

**To Reproduce**
try to find a semantic model permissions using the functions

**Expected behavior**
it should list all item types permessions

**Screenshots**

<img width="1532" height="172" alt="Image" src="https://github.com/user-attachments/assets/0bae799e-9305-4d21-a608-8896ce702948" />

<img width="1584" height="150" alt="Image" src="https://github.com/user-attachments/assets/090ac44a-5583-4998-bef2-4748ffe7a8aa" />
you can see here that type is ommiteed

m-kovalsky:
Thanks. Made a PR (#941) to fix this for the next release.

m-kovalsky:
Fixed in [0.12.7](https://github.com/microsoft/semantic-link-labs/releases/tag/0.12.7).

PR #941, opened 2025-11-02
940

m-kovalsky:
#940

### DIFF
--- a/src/sempy_labs/admin/_items.py
+++ b/src/sempy_labs/admin/_items.py
@@ -234,12 +234,8 @@ def list_item_access_details(
     }
     df = _create_dataframe(columns=columns)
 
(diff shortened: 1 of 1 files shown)


src/sempy_labs/admin/_items.py
```diff
@@ -246,8 +234,12 @@ def list_item_access_details(
     }
     df = _create_dataframe(columns=columns)
 
+    url = f"/v1/admin/workspaces/{workspace_id}/items/{item_id}/users"
+    if type in ['Report', 'Dashboard', 'SemanticModel', 'App', 'Dataflow']:
```
bo-zhou-ms: Can we always put type as a url param? The public doc mentions these types are required to have type param but for others we can pass type for filtering in backend thus having better perf.

bo-zhou-ms (approved):

Issue #950, opened 2025-11-05
Add function to cancel item's job instance


vytautas9:
**Is your feature request related to a problem? Please describe.**
Currently, we have functions for majority of API wrappers of the [Job Scheduler](https://learn.microsoft.com/en-us/rest/api/fabric/core/job-scheduler) endpoints. However at least one is missing to cancel running Job Instances.

We have a use case where we are running a lot of sessions and/or a lot of experiments and want to programmatically cancel them.

**Describe the solution you'd like**
A wrapper function to cancel single job instance. This is the Fabric API that does the exact - [Job Scheduler - Cancel Item Job Instance](https://learn.microsoft.com/en-us/rest/api/fabric/core/job-scheduler/cancel-item-job-instance)

**Describe alternatives you've considered**
not applicable

**Additional context**
I will create a PR with proposed implementation.

m-kovalsky:
See #980. I think that should work. Feel free to test it.

vytautas9:
> See [#980](https://github.com/microsoft/semantic-link-labs/pull/980). I think that should work. Feel free to test it.

Got back from sickness so now had a chance to test, thanks @m-kovalsky !

Hmm it does the job of cancelling the instance but the output is misleading.

Steps to reproduce:

1. Start `Notebook1` with semantic link labs installed
2. Start `Notebook2`
3. Within `Notebook1`, run cancel job instance function on `Notebook2` running instance

Expected behaviour:

- `Notebook2` instance is cancelled
- Output of the function in `Notebook1` indicates the instance has been cancelled

Existing behaviour:
- `Notebook2` instance is cancelled
- Output says `The job instance 'XXXXXX' for the 'Notebook2' notebook has already completed before it could be cancelled.`

This is how the `Notebook2` instance looks after cancelling:

<img width="1614" height="175" alt="Image" src="https://github.com/user-attachments/assets/d5845275-2e3a-4eed-98b6-521a81361f3f" />

As we can see, even though we cancel the spark session through API, the status is showed as `Completed` and therefore the following code snippet is executed after cancelling this session:
```python
elif status == "Completed":
            print(
                f"{icons.info} The job instance '{job_instance_id}' for the '{item_name}' {type.lower()} has already completed before it could be cancelled."
            )
            return
```

m-kovalsky:
But the status shows as Completed. Shouldn't the status show as 'Cancelled'?

vytautas9:
[...]

PR #952, opened 2025-11-05
Add function to cancel item job instance

vytautas9:
Currently, we have functions for majority of API wrappers of the [Job Scheduler](https://learn.microsoft.com/en-us/rest/api/fabric/core/job-scheduler) endpoints. However at least one is missing to cancel running Job Instances. This is the Fabric API that does the exact - [Job Scheduler - Cancel Item Job Instance](https://learn.microsoft.com/en-us/rest/api/fabric/core/job-scheduler/cancel-item-job-instance).

The PR adds a new `cancel_item_job_instance` function to the `src/sempy_labs/_job_scheduler.py` module.

P.S. I had to make sure that **lro** would not be triggered for this endpoint as the endpoint does not return `x-ms-operation-id` and instead returns `x-ms-job-id` + `location` with logic being a bit different than handling the `x-ms-operation-id` case.

### DIFF
--- a/src/sempy_labs/_job_scheduler.py
+++ b/src/sempy_labs/_job_scheduler.py
@@ -12,6 +12,8 @@ from sempy_labs._helper_functions import (
 )
 from uuid import UUID
 import sempy_labs._icons as icons
+import time
+from sempy.fabric.exceptions import FabricHTTPException
 
 
 @log
@@ -539,7 +541,9 @@ def cancel_item_job_instance(
     """
     Cancel an item's job instance.
 
-    This is a wrapper function for the following API: `Job Scheduler - Cancel Item Job Instance <https://learn.microsoft.com/en-us/rest/api/fabric/core/job-scheduler/cancel-item-job-instance>`_.
+    This is a wrapper function for the following API: `Job Scheduler - Cancel Item Job Instance <https://learn.microsoft.com/rest/api/fabric/core/job-scheduler/cancel-item-job-instance>`_.
+
+    Service Principal Authentication is supported (see `here <https://github.com/microsoft/semantic-link-labs/blob/main/notebooks/Service%20Principal.ipynb>`_ for examples).
 
     Parameters
     ----------
@@ -560,15 +564,56 @@ def cancel_item_job_instance(
         item=item, type=type, workspace=workspace
     )
 
-    _base_api(
-        request=f"v1/workspaces/{workspace_id}/items/{item_id}/jobs/instances/{job_instance_id}/cancel",
-        method="post",
-        lro_return_json=False,
-        lro_return_status_code=False,
-        status_codes=[200, 202],
-        client="fabric_sp",
-    )
+    try:
+        response = _base_api(
+            request=f"v1/workspaces/{workspace_id}/items/{item_id}/jobs/instances/{job_instance_id}/cancel",
+            method="post",
+            status_codes=[200, 202],
+            client="fabric_sp",
+        )
+    except FabricHTTPException as e:
+        if (
+            e.status_code == 400
+            and e.response.json().get("errorCode") == "JobAlreadyCompleted"
+        ):
+            # If Job Instance is already completed, skip the rest of the process and output the Job Instance details.
+            print(
+                f"{icons.green_dot} The Job Instance '{job_instance_id}' of '{item_name}' {type.lower()} within the '{workspace_name}' workspace has been already cancelled."
+            )
+
+            return _get_item_job_instance(
+                url=f"v1/workspaces/{workspace_id}/items/{item_id}/jobs/instances/{job_instance_id}"
+            )
+        else:
+            print(
+                f"{icons.red_dot} An error occurred while cancelling '{job_instance_id}' of '{item_name}' {type.lower()} within the '{workspace_name}' workspace: {e}"
+            )
+            raise
 
     print(
-        f"{icons.green_dot} The cancellation of Job Instance '{job_instance_id}' of '{item_name}' {type.lower()} has been started successfully."
+        f"{icons.in_progress} The cancellation of Job Instance '{job_instance_id}' of '{item_name}' {type.lower()} within the '{workspace_name}' workspace has been initiated."
     )
+
+    status_url = response.headers.get("Location").split("fabric.microsoft.com")[1]
+    status = None
+    while status not in ["Completed", "Failed", "Cancelled"]:
+        response = _base_api(request=status_url)
+        status = response.json().get("status")
+        time.sleep(3)
+
+    df = _get_item_job_instance(url=status_url)
+
+    # Check what is the final status of the Job Instance.
+    if status in ["Completed", "Failed", "Cancelled"]:
+        print(
+            f"{icons.green_dot} The Job Instance '{job_instance_id}' of '{item_name}' {type.lower()} within the '{workspace_name}' workspace has been cancelled successfully."
+        )
+    else:
+        print(
+            f"Latest status of Job Instance '{job_instance_id}' of '{item_name}' {type.lower()} within the '{workspace_name}' workspace: {status}"
+        )
+        print(
+            f"{icons.red_dot} The cancellation of Job Instance '{job_instance_id}' of '{item_name}' {type.lower()} within the '{workspace_name}' workspace has failed."
+        )
+
+    return df


vytautas9:
Solves #950

src/sempy_labs/_job_scheduler.py
```diff
@@ -527,3 +527,48 @@ def create_item_schedule_weekly(
...
+def cancel_item_job_instance(
+    item: str | UUID,
+    job_instance_id: UUID,
+    type: Optional[str] = None,
+    workspace: Optional[str | UUID] = None,
+):
+    """
+    Cancel an item's job instance.
+
+    This is a wrapper function for the following API: `Job Scheduler - Cancel Item Job Instance <https://learn.microsoft.com/en-us/rest/api/fabric/core/job-scheduler/cancel-item-job-instance>`_.
```
m-kovalsky: Remove the /en-US part please

m-kovalsky:
One more thing in the review - the function should wait for the cancellation to complete. For this you need to check the status and wait for it to complete. See how it was done here although this one may be slightly different.

https://github.com/microsoft/semantic-link-labs/blob/d8b960ba0078b33a2929af75601ca2d1644c6ce7/src/sempy_labs/lakehouse/_materialized_lake_views.py#L57

src/sempy_labs/_job_scheduler.py
vytautas9: Removed. Also added note on SP authentication support.

vytautas9:
> One more thing in the review - the function should wait for the cancellation to complete. For this you need to check the status and wait for it to complete. See how it was done here although this one may be slightly different.
> 
> https://github.com/microsoft/semantic-link-labs/blob/d8b960ba0078b33a2929af75601ca2d1644c6ce7/src/sempy_labs/lakehouse/_materialized_lake_views.py#L57

Added the following:

- Error handling when the job was already cancelled to not throw exception
- Waiting loop to wait for the job to finish
- Return dataframe of the latest status

vytautas9:
Btw I've noticed a possible issue with `_get_item_job_instance` internal function. It's made to handle multiple responses but if I am not mistaken, the endpoint always return a single response so now the function always return empty dataframe.

As this function is also used in other modules, didn't wanted to try to fix it in the current PR and might test more and create another PR.

m-kovalsky:
> Btw I've noticed a possible issue with `_get_item_job_instance` internal function. It's made to handle multiple responses but if I am not mistaken, the endpoint always return a single response so now the function always return empty dataframe.
> 
> As this function is also used in other modules, didn't wanted to try to fix it in the current PR and might test more and create another PR.

#958 should solve that. Thanks!

src/sempy_labs/_job_scheduler.py
```diff
@@ -527,3 +529,91 @@ def create_item_schedule_weekly(
...
+        Defaults to None which resolves to the workspace of the attached lakehouse
+        or if no lakehouse attached, resolves to the workspace of the notebook.
+    """
+
+    (workspace_name, workspace_id) = resolve_workspace_name_and_id(workspace)
+    (item_name, item_id) = resolve_item_name_and_id(
+        item=item, type=type, workspace=workspace
+    )
+
+    try:
```
m-kovalsky, line 567: the try/except here won't do much because the exception is already handled within the _base_api function.

src/sempy_labs/_job_scheduler.py
```diff
@@ -527,3 +529,91 @@ def create_item_schedule_weekly(
...
+        time.sleep(3)
+
+    df = _get_item_job_instance(url=status_url)
+
+    # Check what is the final status of the Job Instance.
+    if status in ["Completed", "Failed", "Cancelled"]:
+        print(
+            f"{icons.green_dot} The Job Instance '{job_instance_id}' of '{item_name}' {type.lower()} within the '{workspace_name}' workspace has been cancelled successfully."
+        )
+    else:
```
m-kovalsky, line 611: the 'else' portion would never actually be reached because it would still be in the while loop.

src/sempy_labs/_job_scheduler.py
```diff
@@ -527,3 +529,91 @@ def create_item_schedule_weekly(
...
+    status = None
+    while status not in ["Completed", "Failed", "Cancelled"]:
+        response = _base_api(request=status_url)
+        status = response.json().get("status")
+        time.sleep(3)
+
+    df = _get_item_job_instance(url=status_url)
+
+    # Check what is the final status of the Job Instance.
+    if status in ["Completed", "Failed", "Cancelled"]:
```
m-kovalsky, line 607: all 3 of these outcomes mean cancelled successfully? i don't think this is correct.

m-kovalsky (changes requested):

src/sempy_labs/_job_scheduler.py
vytautas9: As far as I know (might be wrong) but we don't have a separate API endpoint to check the status of Cancel operation for Job Instance's. Unlike with Long Running Operations (where we can get the status of the operation), cancelling a Job Instance does not return the operation in it's headers and instead returns only the location of the job instance.

So what I am checking is the status of the Job Instance from https://learn.microsoft.com/en-us/rest/api/fabric/core/job-scheduler/get-item-job-instance?tabs=HTTP (returned also by the headers as location). We wait till the status of it is no longer not started / in progress.

For example, if we cancel an active notebook Spark session: the job instance is marked as "Completed"; if we cancel running pipeline: the job instance is marked as "Cancelled".

Do you know another approach perhaps?

src/sempy_labs/_job_scheduler.py
vytautas9: The try/except will catch the exception thrown by `_base_api` function that is handled internally at `_base_api` and will "silence" the exception if it's `400` and `JobAlreadyCompleted`, if the exception is any other than this, it will reraise it with `raise` in the else statement.

This is how it looks when the `_base_api` returns `400` and `JobAlreadyCompleted` and we handle it:
image  
So the only added benefit of this try/except is to make the code not fail in case we cancel an already finished/cancelled Job Instance.

This is how it looks when the `_base_api` returns an exception that we don't handle:
image  
In this case, the exception is not the one we handle so we re-raise the exception cause by `_base_api` + output additional information to console.



Let me know if you'd like this to be changed or the try/except completely removed. The reasoning behind this was to make the code/notebook not fail if we try to cancel an already finished/cancelled Job Instance (this exception could also be handled by the end users themselves inside code/notebooks).

m-kovalsky: I'd rather not use try/except as it can lead to strange things not being seen properly. What if instead we checked the status of the job and if it is 'Completed' just print an output saying the job is completed and then return. If the job is not completed it can simply run the cancel API step. But the check of the status should not loop. It should be a 1-time check.

src/sempy_labs/_job_scheduler.py
m-kovalsky: If the status shows as 'Cancelled' then the cancel job ran successfully. If the status shows as 'Failed' then the job failed. If the status shows as 'Completed' then the cancel job did not succeed and the job succeeded. Correct?

vytautas9: Not really (as far as I know). The status shows the status of the **Job Instance** itself and not the **cancel operation**. I have not found a way to retrieve the status of the **cancel operation** (not sure if there's one, not like in Long Running Operations. At least didn't found it in the response headers + documented APIs).

For example cancelling running Spark Job Instance (stopping) marks it as "Completed" instead of "Cancelled".

src/sempy_labs/_job_scheduler.py
vytautas9: Ay, can do that. The flow then would be to check the status of the Job Instance before running the cancel API.

Do you think this flow then makes sense?:

1. Check status
2. (if Completed/Failed/Cancelled) Stop and output that the Job is finished and return status.
3. (else (not started / in progress) Run Cancel API
4. Wait few seconds and check status and return
5. (if Job is finished) output the job is cancelled + return status
6. (if Job has not yet finished) output the job cancel is in progress + return status

At the end, as we don't do a loop, if the cancel operation takes a bit longer (and we don't catch it) it is left to end user to implement a wait (if it's needed).

P.S. This complication is only because we can't get information about cancel operation and we don't get `Retry-After` for example.

m-kovalsky:
closing as we have the other PR for this which just went to prod.

PR #953, opened 2025-11-06
M kovalsky/graph model

m-kovalsky:


src/sempy_labs/graph_model/_background_jobs.py
```diff
@@ -0,0 +1,72 @@
...
+    response = _base_api(
+        request=f"/v1/workspaces/{workspace_id}/GraphModels/{item_id}/jobs/instances?jobType=RefreshGraph",
+        method="post",
+    )
+
+    print(
+        f"{icons.in_progress} The refresh graph job for the '{item_name}' graph model within the '{workspace_name}' workspace has been initiated."
+    )
+
+    status_url = response.headers.get("Location").split("fabric.microsoft.com")[1]
```
bo-zhou-ms: Maybe use LRO here?

PR #987, opened 2025-11-27
Enable creating shortcuts pointing to warehouse tables

lBilali:
Fabric supports creating shortcuts pointing to warehouse tables as well
but `create_shortcut_onelake` supports only lakehouse

this PR makes it possible to pass a lakehouse or warehouse name or UUD

for backward compatibility the argument name is still `source_lakehouse` but can be used to pass a lake house or a warehouse

### DIFF
--- a/src/sempy_labs/_helper_functions.py
+++ b/src/sempy_labs/_helper_functions.py
@@ -1749,7 +1749,7 @@ def resolve_warehouse_id(
 
 @log
 def resolve_warehouse_name_and_id(
-    warehouse: str | UUID = None, workspace: Optional[str | UUID] = None
+    warehouse: str | UUID, workspace: Optional[str | UUID] = None
 ) -> Tuple[str, UUID]:
 
     workspace_id = resolve_workspace_id(workspace)


crazy-treyn:
@m-kovalsky FYI https://github.com/microsoft/semantic-link-labs/pull/501 is an open PR from several months back that is ready to merge and will bring support for creating shortcuts to all Fabric data store items:
- Lakehouse
- Warehouse
- SQL Database
- Mirrored Database
- KQL Database

m-kovalsky:
Thanks but I don't think this is necessary now. See #501.

danmeissner:
This change that PR #501 made, limits the OneLake shortcuts to the 5 specified source types.

It does not include the ability to shortcut to the 'MirroredAzureDatabricksCatalog' type.  Which is a valid type ([https://learn.microsoft.com/en-us/rest/api/fabric/core/items/create-item?tabs=HTTP#itemtype](https://learn.microsoft.com/en-us/rest/api/fabric/core/items/create-item?tabs=HTTP#itemtype))

This is important for those of us that are shortcutting directly to Databricks using the built in Mirroring capability.  

The conflict appears to happen when an attempt is made to resolve the item name to an item_id.  If an item_id is used in the create_shortcut_onelake function (not a name) and the source_item_type is not specified, the function will create the shortcuts properly.

Can this additional type be added to the create_shortcut_onelake function so that source_item can either be a name or an ID like it is with the other types?

crazy-treyn:
@danmeissner I submitted https://github.com/microsoft/semantic-link-labs/pull/1022 to address this

PR #988, opened 2025-11-30
Adds Object Display Name to Visual Filter, Page Filter, and Report Filter categories

BeSmarterWithData:
Brings 'Object Display Name' to Visual Filters, Page Filters, and Report Filters.

(this is the remaining 3 categories that allow you to rename a report object in the same way as Visual Objects - where the display name is already captured).

@m-kovalsky This follows the same pattern and logic currently used to pull in the object display name for the Visual Objects.

### DIFF
--- a/src/sempy_labs/report/_reportwrapper.py
+++ b/src/sempy_labs/report/_reportwrapper.py
@@ -790,7 +790,6 @@ class ReportWrapper:
                 hidden = flt.get("isHiddenInViewMode", False)
                 filter_type = flt.get("type", "Basic")
                 filter_used = True if "Where" in flt.get("filter", {}) else False
-                obj_display_name = flt.get("displayName")
 
                 entity_property_pairs = helper.find_entity_property_pairs(flt)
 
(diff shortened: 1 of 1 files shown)


m-kovalsky (changes requested):
thanks. the Object Display Name should default to the Object Name if the Object Default Name is None.

flt.get("displayName", object_name)

BeSmarterWithData:
@m-kovalsky for clarity, do you want that change for the current Visual Objects as well? In the current list_visual_objects, it doesn't fallback to Object Name if there is no alternate display name. 

I will error on the side of consistency and apply to all 4 unless said otherwise :)

m-kovalsky:
> @m-kovalsky for clarity, do you want that change for the current Visual Objects as well? In the current list_visual_objects, it doesn't fallback to Object Name if there is no alternate display name.
> 
> I will error on the side of consistency and apply to all 4 unless said otherwise :)

Yes, good call. Thank you!

BeSmarterWithData:
@m-kovalsky Fallback added for all 4 categories!

m-kovalsky (approved):
Thank you!

PR #992, opened 2025-12-04
Update readme with the Semantic Link pool vote link

ruixinxu:
Update readme with the Semantic Link pool vote link


ruixinxu:
#sign-off

Issue #1001, opened 2025-12-17
Set Endorsement


DomQunis:
Hi Michael and team,

I face the following situation:

I use the feature of cloning Power BI reports to copy reports into other workspaces.
However, after cloning a report I need to set the endorsement (https://learn.microsoft.com/en-us/power-bi/collaborate-share/service-endorse-content#certify-content) manually.

It would be great if I would be able to set endorsement I want automatically by using sempy or a REST API call. 
As far as I know that is not possible with sempy, RESTA PI, or fabric cli.

Therefore, I ask you to provide that feature. Alternatively, I'm open for other suggestions.

Thanks a lot, and keep up your excellent work.

Cheers,
Dom

m-kovalsky:
Would you try PR #1002?

m-kovalsky:
Available in [0.12.9](https://github.com/microsoft/semantic-link-labs/releases/tag/0.12.9).

PR #1002, opened 2025-12-17
M kovalsky/1001

m-kovalsky:
#1001


DomQunis:
Awesome. Happy new year and thanks a lot Michael. Sorry, for my late response. I was alreay on Xmas vacation.

PR #1012, opened 2025-12-25
takeover_item_ownership

m-kovalsky:


src/sempy_labs/_takeover.py
```diff
@@ -0,0 +1,40 @@
...
+from sempy._utils._log import log
+
+
+@log
+def takeover_item_ownership(
+    item: str | UUID,
+    type: Literal["Report", "SemanticModel"],
+    workspace: Optional[str | UUID] = None,
+):
+
```
bo-zhou-ms, line 32: Missing docstring here?

m-kovalsky: thanks. added docstring

PR #1013, opened 2025-12-29
execute_query

m-kovalsky:


src/sempy_labs/dataflow/_items.py
```diff
@@ -0,0 +1,668 @@
...
+@log
+def _resolve_dataflow_name_and_id_and_generation(
+    dataflow: str | UUID, workspace: Optional[str | UUID] = None
+) -> Tuple[str, UUID, str]:
+
+    (workspace_name, workspace_id) = resolve_workspace_name_and_id(workspace)
+
+    dfD = list_dataflows(workspace=workspace_id)
+
+    if _is_valid_uuid(dataflow):
```
bo-zhou-ms, line 275: This approach may fail in scenarios where the the name happens to be a UUID string. While such cases are rare, it's still advisable to account for them.

PR #1056, opened 2026-02-01
operations agents

m-kovalsky:


src/sempy_labs/event_schema_set/_items.py
```diff
@@ -0,0 +1,81 @@
...
+        for v in r.get("value", []):
+            row = {
+                "Event Schema Set Name": v.get("displayName"),
+                "Event Schema Set Id": v.get("id"),
+                "Description": v.get("description"),
+                "OneLake Root Path": v.get("properties", {}).get("oneLakeRootPath"),
+            }
+            rows.append(row)
+
+    return df
```
bo-zhou-ms, line 62: 🔴 Critical: Empty DataFrames Always Returned

**File:** `src/sempy_labs/event_schema_set/_items.py`
**Line:** 58

**File:** `src/sempy_labs/operations_agent/_items.py`
**Line:** 59

Both `list_event_schema_sets` and `list_operations_agents` create an empty dataframe, populate a `rows` list from API responses, but then return the empty dataframe instead of creating a dataframe from the rows list.

**Current pattern:**
```python
df = _create_dataframe(columns=columns)  # Creates empty dataframe

rows = []
for r in responses:
    for v in r.get("value", []):
        rows.append({...})

return df  # Returns EMPTY dataframe, ignoring rows!
```

**Correct pattern (used elsewhere in codebase):**
```python
df = _create_dataframe(columns=columns)

rows = []
for r in responses:
    for v in r.get("value", []):
        rows.append({...})

if rows:
    df = pd.DataFrame(rows, columns=list(columns.keys()))

return df
```

This means **both functions will ALWAYS return empty dataframes** regardless of what the API returns.

bo-zhou-ms (commented):
See comments

src/sempy_labs/event_schema_set/_items.py
m-kovalsky: fixed

PR #1058, opened 2026-02-02
M kovalsky/surge protection

m-kovalsky:


src/sempy_labs/surge_protection/_items.py
```diff
@@ -0,0 +1,353 @@
...
+    resolve_capacity_id,
+    _create_dataframe,
+    _update_dataframe_datatypes,
+)
+import sempy_labs._icons as icons
+from sempy._utils._log import log
+from sempy.fabric.exceptions import FabricHTTPException
+
+
+def _surge_api(
```
bo-zhou-ms, line 16: NIT: just curious, why we stop using the _base_api (like other rest api wrapper) here?

src/sempy_labs/surge_protection/_items.py
```diff
@@ -0,0 +1,353 @@
...
+
+    response = requests.request(
+        method=method,
+        url=f"{prefix}/capacities/{capacity_id}/{request}",
+        headers=headers,
+        json=payload,
+    )
+
+    if response.status_code != status_code:
+        raise FabricHTTPException(response.text)
```
bo-zhou-ms: 🔴 High: Incorrect exception handling in `_surge_api`

**File:** `src/sempy_labs/surge_protection/_items.py`  
**Line:** 30

The helper function passes `response.text` to `FabricHTTPException` instead of the `response` object itself.

**Current:**
```python
raise FabricHTTPException(response.text)
```

**Should be:**
```python
raise FabricHTTPException(response)
```

This is inconsistent with the rest of the codebase and may cause issues with error handling.

src/sempy_labs/surge_protection/_items.py
```diff
@@ -0,0 +1,353 @@
...
+
+    Parameters
+    ----------
+    capacity : str | uuid.UUID, default=None
+        The capacity name or ID.
+        Defaults to None which resolves to the capacity id of the workspace of the attached lakehouse
+        or if no lakehouse attached, resolves to the capacity name of the workspace of the notebook.
+    """
+
+    rules = get_workspace_consumption_rules(capacity=capacity, return_dataframe=False)
```
bo-zhou-ms, line 316: 🔴 High: Potential UnboundLocalError in `delete_workspace_consumption_rules`

**File:** `src/sempy_labs/surge_protection/_items.py`  
**Lines:** 310-320

The function iterates over `rules.get("value", [])` and uses `break` after getting the first `rule_id`. However, if `value` is empty, `rule_id` will be undefined and the subsequent API call will fail with `UnboundLocalError`.

**Current:**
```python
rules = get_workspace_consumption_rules(capacity=capacity, return_dataframe=False)
for v in rules.get("value", []):
    rule_id = v.get("detectionRuleId")
    break
return _surge_api(
    capacity=capacity,
    request=f"detectionRules/{rule_id}",
    ...
)
```

**Suggested fix:**
```python
rules = get_workspace_consumption_rules(capacity=capacity, return_dataframe=False)
value_list = rules.get("value", [])
if not value_list:
    raise ValueError(f"{icons.red_dot} No workspace consumption rules found to delete.")
rule_id = value_list[0].get("detectionRuleId")
```

src/sempy_labs/surge_protection/_items.py
```diff
@@ -0,0 +1,353 @@
...
+        A dictionary showing the background operation surge protection rules for the specified capacity.
+    """
+
+    if rejection_threshold < 1 or rejection_threshold > 100:
+        raise ValueError(
+            f"{icons.red_dot} The rejection_threshold must be between 1 and 100."
+        )
+    if recovery_threshold < 5 or recovery_threshold > 100:
+        raise ValueError(
+            f"{icons.red_dot} The recovery_threshold must be between 1 and 100."
```
bo-zhou-ms: 🟡 Medium: Inconsistent error message in validation

**File:** `src/sempy_labs/surge_protection/_items.py`  
**Line:** 262

The error message says "between 1 and 100" but the validation checks for 5-100:

```python
if recovery_threshold < 5 or recovery_threshold > 100:
    raise ValueError(
        f"{icons.red_dot} The recovery_threshold must be between 1 and 100."
    )
```

Should say "between 5 and 100".

src/sempy_labs/surge_protection/_items.py
m-kovalsky: yes, i will make a PR to fix this.

src/sempy_labs/surge_protection/_items.py
m-kovalsky: fixed

src/sempy_labs/surge_protection/_items.py
m-kovalsky: fixed

src/sempy_labs/surge_protection/_items.py
m-kovalsky: fixed

PR #1059, opened 2026-02-02
enable caching

m-kovalsky:


src/sempy_labs/semantic_model/_caching.py
```diff
@@ -0,0 +1,58 @@
...
+    enable : bool, default=True
+        Set to True to enable query caching, or False to disable it.
+    """
+    prefix = _get_url_prefix()
+    headers = get_pbi_token_headers()
+    (workspace_name, workspace_id) = resolve_workspace_name_and_id(workspace)
+    (item_name, item_id) = resolve_item_name_and_id(
+        item=dataset, type="SemanticModel", workspace=workspace_id
+    )
+    model_id = get_model_id(item_id=item_id, headers=headers, prefix=prefix)
```
bo-zhou-ms, line 40: 🔴 **High:** Missing validation for `get_model_id()` return value.

`get_model_id()` can return `None` if the API response is missing the expected `model.id` structure. This would result in a request to `{prefix}/metadata/models/None/caching`, causing an unclear failure.

**Suggested fix:**
```python
model_id = get_model_id(item_id=item_id, headers=headers, prefix=prefix)
if model_id is None:
    raise ValueError(f"Failed to retrieve model ID for semantic model '{item_name}'")
```

src/sempy_labs/semantic_model/_caching.py
```diff
@@ -0,0 +1,58 @@
...
+from sempy._utils._log import log
+import sempy_labs._icons as icons
+
+
+@log
+def enable_query_caching(
+    dataset: str | UUID, workspace: Optional[str | UUID] = None, enable: bool = True,
+):
+    """
+    Enables or disables `query caching <http://aka.ms/queryCaching>_` for a semantic model.
```
bo-zhou-ms: 🟡 **Minor:** Malformed RST link syntax.

The backtick should come before the underscore:

**Current:**
```python
`query caching <http://aka.ms/queryCaching>_`
```

**Should be:**
```python
`query caching <http://aka.ms/queryCaching>`_
```

bo-zhou-ms (commented):
see comments

src/sempy_labs/semantic_model/_caching.py
m-kovalsky: fixed

src/sempy_labs/semantic_model/_caching.py
m-kovalsky: fixed

PR #1061, opened 2026-02-03
M kovalsky/base api int

m-kovalsky:


src/sempy_labs/_catalog.py
```diff
@@ -101,10 +100,11 @@ def _list_endorsements_or_favorites(type: str = "endorsements"):
...
         payload = {**base_payload, "supportedTypes": supported_types}
 
-        response_list = requests.post(
-            "https://df-msit-scus-redirect.analysis.windows.net/metadata/datahub/V2/artifacts",
-            headers=headers,
-            json=payload,
+        response_list = _base_api(
+            request="metadata/datahub/V2/artifacts",
+            client="interal",
+            method="post",
```
bo-zhou-ms, line 106: **Bug**: Typo - `client="interal"` should be `client="internal"`

src/sempy_labs/_helper_functions.py
```diff
@@ -2231,6 +2231,7 @@ def _base_api(
     lro_return_json: bool = False,
     lro_return_status_code: bool = False,
     lro_return_df: bool = False,
+    headers: Optional[dict] = None,
```
bo-zhou-ms, line 2234: **Note**: The new `headers` parameter is added but not used when `client in ['fabric', 'fabric_sp']`. For the `kusto` client path, headers are passed by caller but this function doesn't use them - the caller builds headers via `_get_kusto_headers()` but they get ignored here.

bo-zhou-ms: This is NIT but better to be checked if the design is as expected.

src/sempy_labs/_catalog.py
m-kovalsky: fixed

src/sempy_labs/_helper_functions.py
m-kovalsky: the headers are used in line 2324 of the _helper_functions.py file.

Issue #886, opened 2025-09-25
Functions using deltalake.DeltaTable break on LH shortcuts to WH tables


philosophicles:
**Describe the bug**

Certain functions in sempy_labs.lakehouse cannot handle lakehouses containing shortcuts to warehouse tables. The problem occurs where the `deltalake.DeltaTable` class is used internally (for pure-python notebooks).

I've found these affected functions:

1. get_lakehouse_tables _(only if the `extended` or `count_rows` flags are set)_
2. get_lakehouse_columns
3. optimize_lakehouse_tables _(affected via `optimize_table()` call, not via earlier call to get_lakehouse_tables())_
4. vacuum_lakehouse_tables _(again, affected via internal `vacuum_table()` call)_

I am not 100% sure if there are others affected as well.

**To Reproduce**

1. Create a warehouse with a table in it
2. Create a lakehouse in the same workspace with a shortcut to the warehouse table
3. In a pure-python notebook, call one of the above functions, referencing this lakehouse either explicitly or implicitly (and not referencing specific tables, where that's an option). 

Error will be raised from the deltalake package:

> DeltaError: Generic DeltaTable error: missing-column: createdTime

By comparison, if the lakehouse only includes shortcuts to other lakehouse tables, or does not include any shortcuts, this error isn't encountered.

**Expected behavior**

I don't think the deltalake package is doing anything wrong here; I can't imagine it's meant to work on Fabric Warehouse delta tables, since the Fabric WH engine expects to solely manage them and allow interaction only via the WH SQL endpoint.

I think the sempy-labs functions should either:

* detect if a LH table is a shortcut to a warehouse and if so, skip the relevant bit of code; or
* catch the `DeltaError` if raised and continue gracefully.

In either case, any relevant outputs (such as pandas dfs) can show a None/null/empty kind of value for these cases. For the `get_` functions, it may help to add an extra boolean field to dataframe outputs that explicitly indicates "this row relates to a shortcut to a warehouse table". Then client code can choose how to handle that data.

I do not think any python exception needs to be raised back to the client code in such cases; WH tables cannot be optimised or vacuumed on demand (AFAIK), and I don't think it's unreasonable to return less-complete results from the `get_` functions in this situation.

philosophicles:
Just wanted to observe that the implementation of #888 that is coming together would mean shortcuts are _never_ in-scope for the optimize function, and probably not the vacuum function either. Thus resolving this bug for those two functions.

[...]

PR #1077, opened 2026-02-13
get_lakehouse_tables fix for row_count

m-kovalsky:
#886


src/sempy_labs/lakehouse/_get_lakehouse_tables.py
```diff
@@ -177,35 +190,44 @@ def get_lakehouse_tables(
         df["Parquet File Guardrail Hit"] = df["Files"] > df["Parquet File Guardrail"]
         df["Row Group Guardrail Hit"] = df["Row Groups"] > df["Row Group Guardrail"]
     if count_rows:
-        df["Row Count"] = df["Row Count"].astype(int)
+        if _pure_python_notebook:
```
bo-zhou-ms: should be `if _pure_python_notebook()` ?

m-kovalsky: fixed, thanks!

bo-zhou-ms (approved):

Issue #1078, opened 2026-02-13
Run_table_maintenance is not working correctly with optimize and V-order parameters


Imre-Vermaatgroep:
**Describe the bug**
When using the sempy_labs.lakehouse.run_table_maintenance function with optimize=True and v_order=True parameters the job successfully completes but the is_v_ordered and _delta_log files show different: v_order=False! The optimize with v-order run_table_maintenance seems to not working properly.

Similar to bug [#789](https://github.com/microsoft/semantic-link-labs/issues/789), which should have been fixed since semantic-link-labs 0.11.3.

**To Reproduce**

1. Run the function run_table_maintenance
2. Set the optimize parameter to true and the v_order parameter to true
3. Run the function is_v_ordered

Resulting output from is_v_ordered is not changing to the stated value set in run_table_maintenance, also the _delta_log file of the table doesn't contain vorder=true, as a matter of fact all the Date modified of the table files remain unchanged.

**Expected behavior**
The lakehouse table should have been optimized and v-order should be enabled.

**Screenshots**

<img width="1419" height="422" alt="Image" src="https://github.com/user-attachments/assets/101868ed-3e5a-42c3-ae6e-9686ddab4b0a" />

**Desktop (please complete the following information):**
 - Microsoft Fabric
 - Runtime 1.3 Spark 3.5 Delta 3.2
 - semantic-link-labs 0.13.1

m-kovalsky:
Found a bug in the function. PR #1080 has the fix for the next release.

m-kovalsky:
Fixed in [0.13.2](https://github.com/microsoft/semantic-link-labs/releases/tag/0.13.2)

PR #1080, opened 2026-02-15
1078

m-kovalsky:
#1078

### DIFF
--- a/src/sempy_labs/lakehouse/_helper.py
+++ b/src/sempy_labs/lakehouse/_helper.py
@@ -67,20 +67,35 @@ def is_v_ordered(
 
         latest_file = os.path.join(delta_log_path, json_files[0])
 
+        value_a = False
+        value_b = False
+
         with open(latest_file, "r") as f:
             all_data = [
                 json.loads(line) for line in f if line.strip()
             ]  # one dict per line
             for data in all_data:
                 if "metaData" in data:
-                    value_a = data.get("metaData", {}).get("configuration", {}).get("delta.parquet.vorder.enabled", False)
-                    value_a = bool(value_a)
+                    value_a = (
+                        str(
+                            data.get("metaData", {})
+                            .get("configuration", {})
+                            .get("delta.parquet.vorder.enabled", False)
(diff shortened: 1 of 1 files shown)


src/sempy_labs/lakehouse/_helper.py
```diff
@@ -73,18 +73,18 @@ def read_vorder_tag(delta_log_path):
...
+                    value_a = bool(value_a)
 
             # If no metaData, fall back to commitInfo
             for data in all_data:
                 if "commitInfo" in data:
-                    tags = data["commitInfo"].get("tags", {})
-                    return tags.get("VORDER", "false").lower() == "true"
+                    value_b = data.get('commitInfo', {}).get('tags', {}).get('VORDER', False)
+                    value_b = bool(value_b)
+            if value_a or value_b:
```
bo-zhou-ms, line 99: Considering claiming vlaue_a and value_b with initial values at around line 69? Otherwise they could be undefined here and cause exception.

src/sempy_labs/lakehouse/_helper.py
```diff
@@ -73,18 +73,18 @@ def read_vorder_tag(delta_log_path):
...
             for data in all_data:
                 if "metaData" in data:
-                    return (
-                        data.get("metaData", {})
-                        .get("configuration", {})
-                        .get("delta.parquet.vorder.enabled", "false")
-                        == "true"
-                    )
+                    value_a = data.get("metaData", {}).get("configuration", {}).get("delta.parquet.vorder.enabled", False)
+                    value_a = bool(value_a)
```
bo-zhou-ms: python converts `bool("False")` to True. We need to manual cast the string to boolean here. Same for `value_b`

Image

m-kovalsky: fixed

src/sempy_labs/lakehouse/_helper.py
m-kovalsky: fixed

bo-zhou-ms (approved):

PR #1103, opened 2026-03-08
admin.*_capacity_tenant_setting_override added function and updated another

tarente:
New function "delete_all_capacity_tenant_setting_override"
    Deletes and returns list of tenant setting overrides that override at the capacities after applying the tenant_setting filter.

    This is a wrapper function for the following APIs: `Tenants - List Capacities Tenant Settings Overrides <https://learn.microsoft.com/rest/api/fabric/admin/tenants/list-capacities-tenant-settings-overrides>`_
    and `Tenants - Delete Capacity Tenant Setting Override <https://learn.microsoft.com/rest/api/fabric/admin/tenants/delete-capacity-tenant-setting-override>`_.

    Service Principal Authentication is supported (see `here <https://github.com/microsoft/semantic-link-labs/blob/main/notebooks/Service%20Principal.ipynb>`_ for examples).

    Parameters
    ----------
    capacity : str | uuid.UUID, default = None
        The capacity name or ID.
        Defaults to None which resolves to showing/deleting all capacities.
    tenant_setting : str, default = None
        The tenant setting name. Example: "TenantSettingForCapacityDelegatedSwitch"
        Defaults to None which resolves to showing/deleting all tenant settings.
    dry_run : bool, default = True
        Show or delete the tenant settings override at the capacities
        Defaults to True which resolves to showing the tenant settings override at the capacities.

    Returns
    -------
    pandas.DataFrame
        A pandas dataframe showing a list of tenant setting overrides that override at the capacities after applying the tenant_setting filter.

-------

Added column "Excluded Security Groups" to function "list_capacity_tenant_settings_overrides".

### DIFF
--- a/src/sempy_labs/admin/__init__.py
+++ b/src/sempy_labs/admin/__init__.py
@@ -43,6 +43,7 @@ from ._capacities import (
 )
 from ._tenant import (
     list_tenant_settings,
+    delete_all_capacity_tenant_setting_override,
     delete_capacity_tenant_setting_override,
     update_tenant_setting,
     update_capacity_tenant_setting_override,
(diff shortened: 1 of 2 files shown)


m-kovalsky:
fyi i will update the function name to be plural ('overrides'). this will go into the next release.

Issue #1008, opened 2025-12-22
Vertipaq Results


m-opatovsky:
**Describe the bug**
Similar to issue [#975 ](https://github.com/microsoft/semantic-link-labs/issues/975), results of the Vertipaq Analyzer with table parameter is being written incorrectly into schema-enabled Lakehouses.

**To Reproduce**
1. Start with notebook with an environment configured with semantic-link-labs 0.12.8. Connect to a schema-enabled lakehouse.
2. Run something like `labs.vertipaq_analyzer(
                dataset=dataset,
                workspace=workspace,
                export='table')`
3. Function will execute without error, but create PQT files under the new schemas "vertipaqanalyzer_tables", "vertipaqanalyzer_models", etc.

**Expected behavior**
I think ideally a parameter to pass the destination schema is needed, but similar to the solution for BPA in issue 975, defaulting to the dbo schema is fine.

**Screenshots**

<img width="1067" height="427" alt="Image" src="https://github.com/user-attachments/assets/ef4bbf20-a6aa-47f9-bea0-a895d9e85a39" />

**Desktop (please complete the following information):**
 - OS: Win 11
 - Browser: Edge
 - Version: 143.0.3650.66

m-kovalsky:
See the updates in [0.14.3](https://github.com/microsoft/semantic-link-labs/releases/tag/0.14.3).

Issue #103, opened 2024-08-28
Vertipaq Export


prathyusha-kamasani:
Hi,

Is it possible to update scripts to Export results without attaching the Lakehouse, please? So we can dynamically mount LH and export data into LH tables. Currently, when I mount with script I get below error, it will be good be able to mount dynamically. Many Thanks! Prathy

ValueError: 🔴 In order to save the Vertipaq Analyzer results, a lakehouse must be attached to the notebook. Please attach a lakehouse to this notebook.

m-kovalsky:
I would like to enable saving to any lakehouse in any workspace however currently it is problematic. Saving the results is not the problem. In order to add the proper 'RunId', I must query the delta table to get the max run Id. Running spark queries cross-workspace is not supported. Thus, using the lakehouse attached to the notebook ensures no problems. Once spark queries can be executed cross-workspace, I'd be happy to update the functionality to support saving to any lakehouse in any workspace.

eisber:
@m-kovalsky  you could use delta-rs to read directly in onelake...

https://blog.fabric.microsoft.com/en-us/blog/delta-rs-introduces-native-support-for-onelake/

m-kovalsky:
Done in [0.14.3](https://github.com/microsoft/semantic-link-labs/releases/tag/0.14.3).

PR #1112, opened 2026-03-17
M kovalsky/vertipaq UI

m-kovalsky:


m-kovalsky:
#1008, #103

src/sempy_labs/_sql.py
```diff
@@ -35,43 +35,43 @@ class ConnectBase:
     def __init__(
         self,
         item: str | UUID,
+        type: Optional[str] = "Warehouse",
```
bo-zhou-ms, line 38: NIT: Not blocking, but I would suggest to stop using `type` for all new parameters from now on as it's a builtin keyword in python. Maybe consider continue to `endpoint_type` or similar.

bo-zhou-ms (approved):

PR #1126, opened 2026-03-27
Add fix_hide_visual_filters: hide visual-level filters from filter pane

KornAlexander:
Adds a report fixer function `fix_hide_visual_filters()` that sets `isHiddenInViewMode=True` on all visual-level filters, keeping only page-level and report-level filters visible.


KornAlexander:

KornAlexander:

PR #1128, opened 2026-03-27
Add add_measure_table: empty calculated table for centralizing measures

KornAlexander:
Adds a semantic model fixer `add_measure_table()` that creates an empty calculated table (_Measures) for centralizing measures in one location.


KornAlexander:

PR #1130, opened 2026-03-27
Add add_calc_group_units: Units calculation group (Thousand/Million)

KornAlexander:
Adds a semantic model fixer `add_calc_group_units()` that creates a Units calculation group with Thousand and Million calculation items for unit conversion.


KornAlexander:

PR #1131, opened 2026-03-27
Add fix_discourage_implicit_measures: set DiscourageImplicitMeasures

KornAlexander:
Adds a semantic model fixer `fix_discourage_implicit_measures()` that sets `DiscourageImplicitMeasures=True` on all tables in the model, a well-known best practice for semantic models.


KornAlexander:

PR #1139, opened 2026-04-07
Tarente/list activity events multiple days

tarente:
Added function list_activity_events_multiple_days:
    """
    Shows a list of audit activity events for a tenant.

    This is a wrapper function for the following API: `Admin - Get Activity Events <https://learn.microsoft.com/rest/api/power-bi/admin/get-activity-events>`_.

    Service Principal Authentication is supported (see `here <https://github.com/microsoft/semantic-link-labs/blob/main/notebooks/Service%20Principal.ipynb>`_ for examples).

    Parameters
    ----------
    start_time : str
        Start date and time of the window for audit event results. Example: "2024-09-25T07:55:00".
    end_time : str
        End date and time of the window for audit event results. Example: "2024-09-25T08:55:00".
    activity_filter : str, default=None
        Filter value for activities. Example: 'viewreport'.
    user_id_filter : str, default=None
        Email address of the user.
    return_dataframe : bool, default=True
        If True the response is a pandas.DataFrame. If False returns a dict. Default True

    Returns
    -------
    pandas.DataFrame | dict
        A pandas dataframe or dict showing a list of audit activity events for a tenant.
    """

Added function execute_in_timeslots:
    """
    Execute a function repeatedly with rate limiting, merging results and
    logging progress.

    - Resolves the target function once from the given namespace.
    - Processes each parameters as a separate call with arguments in parameters_list.
    - Enforces a sliding‑window limit: max_per_slot calls per slot_seconds.
    - Sleeps or resets the window when limits are reached.
    - Merges results using deep_merge and counts rows via count_rows.
    - On error, logs the failure and returns partial results immediately.

    Returns the merged result of calling the function func_name with the parameters in the parameters_list.
    """


tarente (commented):
Applied comments from Copilot and executed black.

tarente:
Missing review to be approved or rejected.

m-kovalsky:
@tarente how about we modify the original function to support multiple days (with a limit). I'm sure copilot can handle doing the change.

tarente:
@m-kovalsky the API restricts to get the data from the last 28 days, and the start_time and end_time needs to be in the same day (UTC).
The function list_activity_events already has a start_time and end_time.
Are you proposing that based on those 2 parameters, to loop for all days between the 2 dates, 1 day at a time since that is a limitation of the API?

I will ensure that if the start_time and end_time are outside the boundaries of the API (older than 28 days or after today) are corrected before starting the loop.

m-kovalsky:
> @m-kovalsky the API restricts to get the data from the last 28 days, and the start_time and end_time needs to be in the same day (UTC). The function list_activity_events already has a start_time and end_time. Are you proposing that based on those 2 parameters, to loop for all days between the 2 dates, 1 day at a time since that is a limitation of the API?
> 
> I will ensure that if the start_time and end_time are outside the boundaries of the API (older than 28 days or after today) are corrected before starting the loop.

Yes, that’s what I’m thinking. But there has to be a limit on number of days - otherwise it can get out of hand. Just would rather keep the number of functions to a minimum.

tarente:
Hi @m-kovalsky ,

Instead of adding a limit on the number of days, how about to limit the number of rows after which the function would stop iterating the days and would return the parcial result?
E.g., for very small Tenants, we may be able to return all data, but not for big Tenants.

Do you have any a limit for the maximum number of rows returned (either for a dataframe or a dict)?
If they should be different, do you have any suggestion?

Thank you.

m-kovalsky:
@tarente Hmm, well the other consideration with dates as the restriction is the number of API calls. If you put the dates to cover a year it would do a ton of API calls which is not exactly a desirable outcome. We could have a limit for both days and rows.

tarente:
@m-kovalsky, both the start_time and end_date are changed so that the API is not called for dates outside the boundaries of itself.
    Automatically handles multi-day ranges by splitting into per-day requests.
    (the API requires start and end times within the same UTC day).
    Clamps the time range to the last 28 days and no later than the current day.

What is missing is the limit to protect for large Tenants and how it should be implemented, per max number of days or max number of rows.

Issue #1185, opened 2026-05-05
Tracking: pbi-fixer Python contributions


KornAlexander:
Tracking issue for the PBI Fixer Python contribution.

Each function below ships as a standalone, independent PR (one source file + one `__init__.py` re-export). Maintainer can accept or decline each PR in isolation.

All PRs branch from latest `main` and carry the `[pbi-fixer]` title prefix.

### Report — Visuals (charts)
- [ ] #1186 `fix_pie_chart` (`report/_Fix_PieChart.py`)
- [ ] #1187 `fix_bar_chart` (`report/_Fix_BarChart.py`)
- [ ] #1188 `fix_column_chart` (`report/_Fix_ColumnChart.py`)
- [ ] #1189 `fix_line_chart` (`report/_Fix_LineChart.py`)
- [ ] #1157 `fix_ibcs_variance` (`report/_Fix_IBCSVariance.py`)

### Report — Layout & page-level
- [ ] #1143 `fix_page_size` (`report/_Fix_PageSize.py`)
- [ ] #1190 `fix_visual_alignment` (`report/_Fix_VisualAlignment.py`)
- [ ] #1144 `fix_hide_visual_filters` (`report/_Fix_HideVisualFilters.py`)
- [ ] #1218 `fix_disable_show_items_no_data` (`report/_Fix_DisableShowItemsNoData.py`)
- [ ] #1191 `fix_remove_unused_custom_visuals` (`report/_Fix_RemoveUnusedCustomVisuals.py`)
- [ ] #1192 `fix_migrate_report_level_measures` (`report/_Fix_MigrateReportLevelMeasures.py`)

### Semantic model — Object names
- [ ] #1193 `fix_capitalize_object_names` (`semantic_model/_Fix_CapitalizeObjectNames.py`)
- [ ] #1194 `fix_trim_object_names` (`semantic_model/_Fix_TrimObjectNames.py`)

### Semantic model — Column metadata
- [ ] #1195 `fix_data_category` (`semantic_model/_Fix_DataCategory.py`)
- [ ] #1196 `fix_date_column_format` (`semantic_model/_Fix_DateColumnFormat.py`)
- [ ] #1197 `fix_month_column_format` (`semantic_model/_Fix_MonthColumnFormat.py`)
- [ ] #1198 `fix_sort_month_column` (`semantic_model/_Fix_SortMonthColumn.py`)
[...]

PR #1199, opened 2026-05-05
[pbi-fixer] Add fix_flag_column_format: format integer flag columns as Yes/No

KornAlexander:
Adds a sempy_labs fixer `fix_flag_column_format()` that format integer flag columns as Yes/No.

Adds **one** source file (`src/sempy_labs/semantic_model/_Fix_FlagColumnFormat.py`) plus one re-export line in the matching `__init__.py`. No other files touched, no shared helpers, no dependency on any other PR.

**Part of the PBI Fixer contribution.** Tracking issue: #1185


m-kovalsky:
This does not fix the issue as the column is a Boolean and would need to be changed upstream to a Yes/No string.

PR #1219, opened 2026-05-05
[pbi-fixer] Add fix_measure_format: set #,0 format on measures with no format string

KornAlexander:
Adds a sempy_labs fixer `fix_measure_format()` that set #,0 format on measures with no format string.

Adds **one** source file (`src/sempy_labs/semantic_model/_Fix_MeasureFormat.py`) plus one re-export line in the matching `__init__.py`. No other files touched, no shared helpers, no dependency on any other PR.

**Part of the PBI Fixer contribution.** Tracking issue: #1185


m-kovalsky:
This issue cannot be fixed simply as this may not be the correct formatting. We can raise the issue in bpa but it is up to the dev to set the correct formatting.

PR #1246, opened 2026-05-25
Add DAX query timing capture helper

edwardpcharles:
Adds a new public sempy_labs API, capture_dax_query_timings, that runs a DAX query under SSAS trace and returns a single summary DataFrame with total, storage engine, and formula engine timings. Also wires the function into package exports and adds targeted unit tests.

### DIFF
--- a/tests/test_capture_dax_query_timings.py
+++ b/tests/test_capture_dax_query_timings.py
@@ -58,10 +58,16 @@ def test_capture_dax_query_timings_returns_summary_df(monkeypatch):
         ]
     )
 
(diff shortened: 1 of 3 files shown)


tests/test_capture_dax_query_timings.py
```diff
@@ -0,0 +1,161 @@
...
+        return self
+
+    def __exit__(self, exc_type, exc, tb):
+        return False
+
+    def create_trace(self, event_schema):
+        return _FakeTrace(self._df)
+
+
+def test_capture_dax_query_timings_returns_summary_df(monkeypatch):
```
m-kovalsky: Test should be in the src/tests folder

Issue #1270, opened 2026-06-18
Regression in v0.15+: generate_direct_lake_semantic_model() no longer supports overwrite, preventing in-place semantic model updates


Vera-OT:
## Describe the bug

Starting with Semantic Link Labs v0.15.x, the `overwrite` parameter appears to have been removed from `directlake.generate_direct_lake_semantic_model()`.

In v0.14.3 and earlier, we successfully used:

```
directlake.generate_direct_lake_semantic_model(
    dataset=model_name,
    lakehouse_tables=table_names,
    lakehouse=lakehouse_name,
    workspace=workspace_name,
    overwrite=True
)
```

to regenerate an existing Direct Lake semantic model while preserving the semantic model identity and existing report bindings.

After upgrading to v0.15.x, the `overwrite` parameter is no longer available, which prevents us from updating an existing semantic model through the same workflow. We have to create a new semantic model and rebind reports.

## To Reproduce

1. Install or use Semantic Link Labs v0.15.x or later.
2. Attempt to execute:

```
directlake.generate_direct_lake_semantic_model(
    dataset=model_name,
    tables=table_names,
    source=lakehouse_name,
    source_type="Lakehouse",
    source_workspace=workspace_name, 
    overwrite=True
)
```

3. Observe that the function no longer accepts the `overwrite` argument.
```
TypeError
got an unexpected keyword argument 'overwrite'
```

## Expected behavior

Keep the 'overwrite' argument available.

## Screenshots

v0.14.3
<img width="1383" height="908" alt="Image" src="https://github.com/user-attachments/assets/57f75e5b-5c67-4269-b12a-85e18abb99ff" />

Latest

<img width="1647" height="979" alt="Image" src="https://github.com/user-attachments/assets/22c1658f-42f4-4ee0-a0a9-51136982530e" />

## Desktop (please complete the following information)

* OS: Windows 11, Microsoft Fabric Notebook Runtime
* Browser: Edge/Chrome
* Semantic Link Labs version:

  * Working: v0.14.3
  * Issue observed: v0.15.x and later

## Additional context

If this functionality (i.e., overwrite) has intentionally been removed, could you please advise on the recommended approach for performing a full semantic model replacement/update while preserving the existing semantic model and report bindings? Thank you!

m-kovalsky:
Fixed in [0.16.0](https://github.com/microsoft/semantic-link-labs/releases/tag/0.16.0).

PR #1279, opened 2026-07-13
1270

m-kovalsky:
#1270

### DIFF
--- a/src/sempy_labs/directlake/_dl_helper.py
+++ b/src/sempy_labs/directlake/_dl_helper.py
@@ -143,4 +143,5 @@ def generate_direct_lake_semantic_model(
         workspace=workspace,
         refresh=refresh,
         inherit_descriptions=inherit_descriptions,
+        overwrite=overwrite,
     )
--- a/src/sempy_labs/directlake/_dl_helper.py
+++ b/src/sempy_labs/directlake/_dl_helper.py
@@ -91,6 +91,7 @@ def generate_direct_lake_semantic_model(
     workspace: Optional[str | UUID] = None,
     refresh: bool = True,
     inherit_descriptions: bool = True,
+    overwrite: bool = False,
 ):
     """
     Dynamically generates a Direct Lake semantic model based on tables in Fabric.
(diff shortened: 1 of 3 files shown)


src/sempy_labs/semantic_model/_generate.py
```diff
@@ -197,18 +200,17 @@ def generate_direct_lake_semantic_model(
...
-    dfD_filt = dfD[dfD["Dataset Name"] == dataset]
-
-    if dfD_filt.empty:
+    try:
+        dataset_id = resolve_item_id(item=dataset, type="SemanticModel", workspace=workspace_id)
+        if not overwrite:
+            raise ValueError(
+                f"{icons.red_dot} A dataset with the name '{dataset}' already exists in the workspace '{workspace_name}'. Please choose a different name or set 'overwrite=True' to overwrite the existing dataset."
+            )
+    except:
```
whiskyboy: 🔴 **Critical:** `overwrite=False` no longer protects an existing semantic model here.

When the dataset already exists, `resolve_item_id(...)` succeeds, then the explicit `ValueError` for `not overwrite` is raised, but this bare `except:` immediately catches it and calls `create_blank_semantic_model(...)`. Since that helper defaults to `overwrite=True`, the default path can replace an existing model unexpectedly. The broad `except` also hides unrelated resolution errors.

**Suggested fix:** avoid the broad catch and pass the flag through to the existing helper, e.g.:

```python
dataset_id = create_blank_semantic_model(
    dataset=dataset,
    workspace=workspace_id,
    overwrite=overwrite,
)
```

src/sempy_labs/directlake/_dl_helper.py
```diff
@@ -91,6 +91,7 @@ def generate_direct_lake_semantic_model(
     workspace: Optional[str | UUID] = None,
     refresh: bool = True,
     inherit_descriptions: bool = True,
+    overwrite: bool = False,
```
whiskyboy, line 94: 🔴 **High:** The new public `overwrite` parameter is never forwarded to the implementation.

Callers using `sempy_labs.directlake.generate_direct_lake_semantic_model(..., overwrite=True)` currently get the implementation default instead of the value they passed. Please pass it through to `gen(...)`:

```python
gen(
    ...
    inherit_descriptions=inherit_descriptions,
    overwrite=overwrite,
)
```

m-kovalsky: fixed

whiskyboy (approved):
LGTM. The previously blocking overwrite issues are fixed.

PR #1280, opened 2026-07-13
new funcs - gateway, onelake

m-kovalsky:


src/sempy_labs/__init__.py
```diff
@@ -339,11 +339,6 @@
     update_sql_audit_settings,
     set_audit_actions_and_group,
 )
-from ._onelake import (
-    get_onelake_settings,
```
whiskyboy, line 343: 🔴 **High:** This removes existing top-level public exports for the OneLake functions.

Existing users can currently call `sempy_labs.get_onelake_settings()`, `sempy_labs.modify_onelake_diagnostics()`, and `sempy_labs.modify_immutability_policy()`. Moving these into `sempy_labs.onelake` is fine as an additive path, but removing the top-level aliases breaks backward compatibility, which the repo instructions explicitly require.

**Suggested fix:** keep re-exporting these functions from `src/sempy_labs/__init__.py`, importing them from the new `onelake` module instead of `._onelake`.

m-kovalsky: fixed

whiskyboy (approved):
Approved.

PR #1287, opened 2026-07-18
Fix broken unit math in get_semantic_model_size()

### LINKED ISSUE #1284: Suspicious unit conversion in get_semantic_model_size()
**Describe the bug**
Some bizarre math is going on in [get_semantic_model_size()](https://github.com/microsoft/semantic-link-labs/blob/3af2149f7a1de8d7ddc0102eb08fd67f55fe0c19/src/sempy_labs/_generate_semantic_model.py#L615) with KB to 1000 conversion (or vice versa — hard to say).

The docstring clearly says:

```
    Returns
    -------
    int
        The size of the semantic model in bytes
```

Since `model_size` is a summation of integer value columns, it is highly likely that the values represent bytes, not a formatted "display size" in KB, MB, etc. Moreover, if the result is coming from individual values such as

500 — would-be bytes
500 — would-be bytes

Added together, 1000 bytes will be divided by 1024 to produce... what exactly? 0.976KiB? Then multiplication by 1000 gives what unit? Ki*B/k? What does that physically mean?

Even if individual values in `INFO.STORAGETABLECOLUMNS()` are known to misbehave around 1024 factor (proof?), how to reconcile that with summation? 499 + 500 will return 999, but 500 + 500 will return 976.5625, violating the expectation of monotonic increase of the result with increase of contributing components.

The final nail is

```
    return float(result)
```

How does it align with the documented return value? Size in bytes can be only integer.

**Expected behavior**
Keep the original units of the size measurement, that are presumably already bytes.

**Additional context**
If the math somehow makes sense in a way that I cannot comprehend, please leave a comment.

wakinniranye31:
Closes #1284.

## Bug

`get_semantic_model_size()` sums `DICTIONARY_SIZE` and `USED_SIZE` (both already in bytes, from `INFO.STORAGETABLECOLUMNS()` / `INFO.STORAGETABLECOLUMNSEGMENTS()`), then ran that sum through a rescale step keyed off which size bracket it landed in:

```python
if model_size >= 10**9:
    result = model_size / (1024**3) * 10**9
elif model_size >= 10**6:
    result = model_size / (1024**2) * 10**6
elif model_size >= 10**3:
    result = model_size / (1024) * 10**3
else:
    result = model_size
return float(result)
```

As pointed out in the issue, this breaks monotonicity right at each bracket boundary:

```python
>>> old_buggy(999)
999.0
>>> old_buggy(1000)   # more bytes in, smaller number out
976.5625
```

It also returns a `float`, even though the docstring documents `int` / "size in bytes".

## Fix

Removed the rescaling block entirely and return the raw byte sum as an `int`:

```python
dict_size = dict["[DICTIONARY_SIZE]"].sum()
used_size = used_size["[USED_SIZE]"].sum()
model_size = dict_size + used_size

return int(model_size)
```

Also added a `-> int` return annotation to match the docstring.

## Testing

Added `tests/test_get_semantic_model_size.py`, following the same `monkeypatch.setattr(module.fabric, "evaluate_dax", ...)` pattern already used in `tests/test_dax.py`:

- `test_get_semantic_model_size_returns_raw_byte_sum` — 500 + 500 → `1000` (`int`), no rescaling.
- `test_get_semantic_model_size_is_monotonic` — reproduces the exact regression from the issue: 499+500 → 999, 500+500 → 1000, and asserts `1000 > 999` (this would fail against the old code, since 500+500 previously produced 976.5625).

Ran locally:
- `pytest -s tests/` — 7 passed (5 pre-existing + 2 new), no regressions.
- `black --check` / `flake8` on both changed files — clean (confirmed the only `black`/`flake8` deltas elsewhere in the file are pre-existing and untouched by this change).

### DIFF
--- a/src/sempy_labs/_generate_semantic_model.py
+++ b/src/sempy_labs/_generate_semantic_model.py
@@ -614,7 +614,7 @@ def get_semantic_model_definition(
 @log
 def get_semantic_model_size(
     dataset: str | UUID, workspace: Optional[str | UUID] = None
-):
+) -> int:
     """
     Gets size of the semantic model in bytes.
 
@@ -651,14 +651,5 @@ def get_semantic_model_size(
     dict_size = dict["[DICTIONARY_SIZE]"].sum()
     used_size = used_size["[USED_SIZE]"].sum()
     model_size = dict_size + used_size
-    # Calculate proper bytes size by dividing by 1024 and multiplying by 1000 - per 1000
-    if model_size >= 10**9:
-        result = model_size / (1024**3) * 10**9
-    elif model_size >= 10**6:
-        result = model_size / (1024**2) * 10**6
-    elif model_size >= 10**3:
-        result = model_size / (1024) * 10**3
-    else:
-        result = model_size
 
-    return float(result)
+    return int(model_size)
--- a/tests/test_get_semantic_model_size.py
+++ b/tests/test_get_semantic_model_size.py
@@ -1,0 +1,50 @@
+import pandas as pd
+
+import sempy_labs._generate_semantic_model as generate_semantic_model
+from sempy_labs import get_semantic_model_size
+
+
+def _mock_evaluate_dax_factory(dictionary_size_values, used_size_values):
+    def _mock_evaluate_dax(dataset, workspace, dax_string):
+        if "DICTIONARY_SIZE" in dax_string:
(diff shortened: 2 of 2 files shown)


greatvovan:
For awareness, https://github.com/microsoft/semantic-link-labs/issues/1284#issuecomment-5029104389.

Issue #1111, opened 2026-03-17
optimize_lakehouse_tables and vacuum_lakehouse_tables do not work


mrbartuss:
**Describe the bug**
The [optimize_lakehouse_tables](https://semantic-link-labs.readthedocs.io/en/latest/sempy_labs.lakehouse.html#sempy_labs.lakehouse.optimize_lakehouse_tables) and [vacuum_lakehouse_tables](https://semantic-link-labs.readthedocs.io/en/latest/sempy_labs.lakehouse.html#sempy_labs.lakehouse.vacuum_lakehouse_tables) functions from sempy_labs.lakehouse report completion in 4-5 seconds, but the loading bar freezes midway, and no changes occur in the table file view. Manual table maintenance via run_table_maintenance or direct Delta commands succeeds.

**To Reproduce**
I've tried to run in either for one table, list of tables or None - neither worked.

m-kovalsky:
Did you run this in a pure python or PySpark notebook? Would you share more details (i.e. error message in text)? I tested this but it ran successfully in both cases (python & pyspark).

mrbartuss:
I used pure python. I’ll share more details later. Did you run it on a schema-enabled lakehouse?

m-kovalsky:
Yes. For a schema-enabled lakehouse you need to specify the table in the format ‘schema.table’:

tables=['dbo.FactSales']

mrbartuss:
Here is the code I ran:

```
%pip install semantic-link-labs -q
import sempy_labs.lakehouse as lake

lakehouse = None # Enter the name or ID of the lakehouse
workspace = None # Enter the name or ID of the workspace in which the lakehouse exists

df = lake.get_lakehouse_tables(lakehouse=lakehouse, workspace=workspace)

tables = [f"{schema}.{table}" for schema, table in zip(df['Schema Name'], df['Table Name'])]
lake.vacuum_lakehouse_tables(tables=tables, lakehouse=lakehouse, workspace=workspace)
```


And here is the result
<img width="870" height="250" alt="Image" src="https://github.com/user-attachments/assets/3cb2cbef-27f4-4092-acd0-37cd694a05fc" />

When I check the tables file view, there are still many files from the past
<img width="1101" height="184" alt="Image" src="https://github.com/user-attachments/assets/1c02ac6b-bbe0-4ec0-92aa-3a3e9886d864" />

m-kovalsky:
[...]

PR #1288, opened 2026-07-19
1284

m-kovalsky:
#1284


m-kovalsky:
#1111

src/sempy_labs/_generate_semantic_model.py
```diff
@@ -630,35 +630,26 @@ def get_semantic_model_size(
...
-    elif model_size >= 10**6:
-        result = model_size / (1024**2) * 10**6
-    elif model_size >= 10**3:
-        result = model_size / (1024) * 10**3
-    else:
-        result = model_size
+    result = fabric.evaluate_dax(dataset=dataset, workspace=workspace, dax_string=dax)
 
-    return float(result)
+    return int(float(result.iloc[0, 0]))
```
greatvovan, line 655: Just curious, why `int(float(...))`? What's the original type returned by `result.iloc[0, 0]`?

greatvovan:
For awareness: https://github.com/microsoft/semantic-link-labs/issues/1284#issuecomment-5029104389

src/sempy_labs/_generate_semantic_model.py
m-kovalsky: when in a pyspark notebook, this caused issues unless i cast it as a float first.

src/sempy_labs/lakehouse/_lakehouse.py
```diff
@@ -256,11 +233,17 @@ def vacuum_lakehouse_tables(
...
         full_table_name = f"{schema_name}.{table_name}"
-        path = r["Location"]
         bar.set_description(
             f"Vacuuming the '{full_table_name if schema_name else table_name}' table ({idx + 1}/{total})..."
         )
-        _vacuum_table(path=path, retain_n_hours=retain_n_hours)
+        run_table_maintenance(
+            table_name=table_name,
+            vacuum=True,
+            retention_period=f"{retain_n_hours // 24}:{retain_n_hours % 24}:00:00"
```
whiskyboy: missing a "," here?

whiskyboy (approved):

whiskyboy (approved):

src/sempy_labs/_generate_semantic_model.py
greatvovan: My tests show that the original type is `numpy.int64`.  This should not cause issues converting into Python `int` (in fact, it can hold more). Did you investigate it? Maybe it was some edge case conversion of `NaN` or something like that, which we should handle?

PR #1324, opened 2026-08-18
Add delete_graphql_api function

iurii-iurchenko-1:
Adds a dedicated delete function for the GraphQLApi item type, wrapping [Items - Delete GraphQLApi](https://learn.microsoft.com/rest/api/fabric/graphqlapi/items/delete-graphqlapi).

The `graphql` submodule currently covers only `list` and `create` out of the 7 GraphQLApi item APIs. This adds `delete`, following the existing `delete_warehouse_snapshot` / `delete_variable_library` pattern via the shared `delete_item` helper.

**Changes**
- `graphql/_items.py` — added `delete_graphql_api`
- `graphql/__init__.py` — exported it
- `_graphQL.py` — top-level wrapper, matching the two existing GraphQL functions
- `__init__.py` — exported at the package level

**Testing**
Built the wheel and ran it in a Fabric notebook against a live workspace:

```python
labs.create_graphql_api(name="test_delete_me", workspace=ws)
display(labs.list_graphql_apis(workspace=ws))

graphql.delete_graphql_api(graphql_api="test_delete_me", workspace=ws)
display(labs.list_graphql_apis(workspace=ws))
```

The API was created, appeared in the list, was deleted successfully, and no longer appeared in the subsequent list.

Formatted with `black==25.1.0`.

Happy to follow up with `get`, `update`, and the get/update definition APIs for GraphQLApi if that would be useful.

### DIFF
--- a/src/sempy_labs/_graphQL.py
+++ b/src/sempy_labs/_graphQL.py
@@ -52,3 +52,25 @@ def create_graphql_api(
     """
 
     graphql.create_graphql_api(name=name, description=description, workspace=workspace)
+
+
+@log
+def delete_graphql_api(graphql_api: str | UUID, workspace: Optional[str | UUID] = None):
+    """
+    Deletes a GraphQL API.
+
+    This is a wrapper function for the following API: `Items - Delete GraphQLApi <https://learn.microsoft.com/rest/api/fabric/graphqlapi/items/delete-graphqlapi>`_.
+
+    Service Principal Authentication is supported (see `here <https://github.com/microsoft/semantic-link-labs/blob/main/notebooks/Service%20Principal.ipynb>`_ for examples).
+
+    Parameters
+    ----------
+    graphql_api : str | uuid.UUID
+        Name or ID of the GraphQL API.
+    workspace : str | uuid.UUID, default=None
+        The Fabric workspace name or ID.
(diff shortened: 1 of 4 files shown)


iurii-iurchenko-1:

src/sempy_labs/_graphQL.py
```diff
@@ -52,3 +52,25 @@ def create_graphql_api(
     """
 
     graphql.create_graphql_api(name=name, description=description, workspace=workspace)
+
+
+@log
+def delete_graphql_api(graphql_api: str | UUID, workspace: Optional[str | UUID] = None):
```
m-kovalsky: the sempy_labs._graphQL content will be removed in the future (instead use sempy_labs.graphql) so please remove this.

m-kovalsky (changes requested):

src/sempy_labs/_graphQL.py
iurii-iurchenko-1: Removed. 
Also reverted the top-level export in __init__.py, since it imported from _graphQL and would have broken import sempy_labs. delete_graphql_api now lives only in sempy_labs.graphql. Re-tested in a Fabric notebook - created and deleted a GraphQL API by name and by ID.

m-kovalsky (approved):

PR #1334, opened 2026-09-10
Add expression_name parameter for keeping the source name as a part of the expression_name

lucklaytman:
Added an optional expression_name parameter to allow  idempotent updates for multi-source models


lucklaytman:
@microsoft-github-policy-service agree [company=default"]

